PR #3, opened 2015-10-11
Cleanup onion skin toggle

le717:
Hi. :)

This PR does multiple things, all related to onion skinning, but also can help in general development.

First, the general changes:
* Fixes various JS warnings and errors
* Fixes various invalid CSS and HTML
* Removes all excess whitespace (helps make it all a bit neater)
* Adds an `app/variables.js` script. What you can do is define commonly accessed DOM elements here then reference them via a function call, so if/when the selector is changed, you have less required places to change it.

Now for the onion skin changes:
* Store the on/off state as a boolean value in the code rather than changing the IDs. The DOM is not the place to keep such data, and changing the IDs led to some seriously funky conditions.
* Offload the toggle button styling to the CSS where it belongs. I also made it where the button stays the same size no matter the text.
* Make the options panel toggle nicer by:
  * Adding a simple sliding effect
  * Toggling the visibility by toggling a class instead of the ID
* Do the same class toggling action for the framing window
* Improve the IDs for each element affected.

I just noticed I missed a spot that is affected by the refactoring and that you have been working at the same time I have. I'll fix those right up and let you know when I'm done. :)


le717:
All done. Ready for you to take a look and hopefully merge. :)

PR #6, opened 2015-10-12
Improve app loading + new frame rate input

le717:
By moving the JS loading to right before the closing `</body>` tag (as is the universally recommended method), all selector loading can be initialized right then instead of inside `startup()`. :smile: 

Also reworked the frame rate input to be an `input type="number"` field instead of a `confirm` dialog, which is a much nicer UX. :)

If you have any questions, comments, or concerns on any of this, please, do not hesitate to ask! :smiley:

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -42,8 +42,8 @@
                 <h2><img src="icons/play.png"/><span>Playback</span></h2>
                 <ul>
                     <li>
-                      <label id="label-fps-change" for="input-fps-change">Select frame rate</label>
-                      <input id="input-fps-change" type="number" value="15">
+                      <label id="label-fr-change" for="input-fr-change">Select frame rate</label>
+                      <input id="input-fr-change" type="number" value="15">
                     </li>
                     <li>
                       <input id="loopCheckbox" type="checkbox" name="options" value="Download">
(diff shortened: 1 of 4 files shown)


app/index.html
```diff
@@ -132,8 +135,8 @@ <h2 id="playbackTitle">Playback</h2>
         </div>
     </div>
 
-
-
-
+    <script src="variables.js"></script>
```
le717: `type="text/javascript"` is no longer needed with HTML5.

app/index.html
```diff
@@ -44,8 +41,14 @@
...
                 <ul>
-                    <li><a id="changeFrameRate" href="#"> Select frame rate</a></li>
-                    <li><input id="loopCheckbox" type="checkbox" name="options" value="Download"/><span> Loop frames on playback</span></li>
+                    <li>
+                      <label id="label-fr-change" for="input-fr-change">Select frame rate</label>
+                      <input id="input-fr-change" type="number" value="15">
+                    </li>
+                    <li>
+                      <input id="loopCheckbox" type="checkbox" name="options" value="Download">
+                      <label for="loopCheckbox"><span>Loop frames on playback</span></label>
```
le717: I gave the checkbox a `<label>` so users are not forced to click the check, but now can click the text and it will work (also a better UX).

PR #9, opened 2015-10-12
Set the image size in the HTML

le717:
There's actually a difference between setting it in the HTML (the preferred way) VS setting it in CSS. I know this isn't all of them, but I was focusing on this area so that's all I fixed.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -21,7 +21,7 @@
     <div id="sidebar">
         <ul>
             <li>
-                <h2><img src="icons/onionskin.png"/><span>Onion Skinning</span>
+                <h2><img width="15" height="15" src="icons/onionskin.png"><span>Onion Skinning</span>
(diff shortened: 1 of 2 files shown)


BoatsAreRockable:
I've merged the commit but I was wondering what the advantage of setting img sizes via html rather than CSS is? Is it performance related?

Thanks

le717:
There is no real definitive answer for HTML over CSS, but it mostly comes down to
* Spec recommendation
* Site performance
* Browser parsing
* Avoid possible responsive design issues

There is some excellent discussion on the topic [here](http://stackoverflow.com/questions/640190/image-width-height-as-an-attribute-or-in-css) and [here](http://stackoverflow.com/questions/1247685/should-i-specify-height-and-width-attributes-for-my-imgs-in-html).

BoatsAreRockable:
Alright I'll give those a read :)

PR #11, opened 2015-10-14
More fixes

le717:
I wanted to do more, but I'm on mobile an away from my code, so this is all I have right now.


app/main.js
```diff
@@ -469,37 +474,70 @@ function onionSkinAmount() {
...
+}
+
+/**
+ * Display the frame destination directory in the UI.
+ *
+ * @param {String} dir The directory to display.
+ */
+function _displayDirectory(dir) {
+    console.log(`Current destination directory is ${frameExportDirectory}`);
+    document.getElementById("currentDirectoryName").innerHTML = dir;
```
le717: Ugh, this will bug. I'll fix it ASAP.

le717:
PR ready to merge. In this one, I dropped the variables.js script I added (it is not totally needed), documented and better named the file delete code, simplified the date values, and refactored the default save path code (this will help when proper projects are implemented).

app/main.js
```diff
@@ -469,37 +474,70 @@ function onionSkinAmount() {
...
+
+/**
+ * Display the frame destination directory in the UI.
+ *
+ * @param {String} dir The directory to display.
+ */
+function _displayDirectory(dir) {
+    console.log(`Current destination directory is ${dir}`);
+    document.getElementById("currentDirectoryName").innerHTML = dir;
+    document.title = `Boats Animator (${dir})`;
```
le717: This little bit right here is an [ES6 template string](http://davidwalsh.name/javascript-template-strings). They are really cool but sadly can't used full-time on the web. However, because BA is running in an NW.js wrapper that uses the latest Blink and V8 engines that power Chrome, we are use them all we want. Cool, huh? :smiley:

PR #12, opened 2015-10-15
Sliding sidebar and layout fixes

le717:
~~Do not merge this yet as it is still WIP, but I am putting it up already because I want you to be aware of it and the many CSS changes that took 2 hours to perform that cannot be quickly and easily redone and merged with any uncommitted changes you may have.~~

~~I will complete this PR and fully document it after I get some sleep. :)~~

This pull request converts the sidebar into a sliding sidebar, visible by default, with a temporary toggle button. I do not know where the button would go (ideally under View > Toggle Sidebar when there are native menus) so I left it like this.

By nature, this type of sidebar covers the main UI, so this was an opportunity to expand the UI to take up the newly available space. However, doing so was proving quite difficult until I realized the CSS [`box-sizing`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) property was not applied. Despite how scary that page may sound, `box-sizing: border-box;` is actually a [good](https://css-tricks.com/box-sizing/) [thing](http://blog.teamtreehouse.com/box-sizing-secret-simple-css-layouts) when it comes to design modern websites and web apps and makes designs better and easier to work with.

However, this cannot be retroactively applied and things still work. Therefore, I had to rework some values to make everything work. The result? Everything looks a tad nicer. There are a few small visual bugs but nothing major.

I recorded a gif to show off the sidebar and UI tweaks: https://dl.dropboxusercontent.com/u/58490278/Other/boats.gif

Remember, if you have any questions, concerns, or comments, please ask/state them. :)


BoatsAreRockable:
This conflicts with the commit I made 5 minutes ago, sorry :( 

Would you be able to resolve it?
Thanks

BoatsAreRockable:
Thanks for your continued excellent work!

I wonder if it would be better for the main UI to shrink in width when the toggle is activated rather than being covered - this would allow users the option of keeping the sidebar open all of the time rather than being forced to close it to use the playback options. Although this does of course make it less clear the sidebar can be collapsed

Also to make the sidebar easily discoverable I think it would be best if it was open by default.

In terms of where the button to toggle the sidebar would be I definitely think it  should be in `View > Toggle Sidebar` but in addition there should be a button somewhere in the UI.

le717:
> This conflicts with the commit I made 5 minutes ago, sorry :(

Would you be able to resolve it?

I think I can. Give me a few minutes. I'll reply to your other comment afterward. ;)

le717:
Updated and able to be merged.

> I wonder if it would be better for the main UI to shrink in width when the toggle is activated rather than being covered - this would allow users the option of keeping the sidebar open all of the time rather than being forced to close it to use the playback options. Although this does of course make it less clear the sidebar can be collapsed

I considered this but decided against it as it growing/shrinking a layout is really hard and leads to hacky code.

> Also to make the sidebar easily discoverable I think it would be best if it was open by default.

Which it is, as you can see in the gif.

> there should be a button somewhere in the UI.

But where I have no clue. The one I added literally says "Temporary Sidebar toggle". :P

PR #16, opened 2015-10-15
Various JS fixes

le717:
Lots of stuff in here! :smile: 

* Switch a lot of selectors to [`querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)
* Remove unneeded vendor-prefixed code
* Remove old code and blank spaces
* Remove duplicate frame rate value
* Clean up news feed attributes (all were unneeded)
* Clean up the event listeners
* Consistently use double quotes in most places (they were the predominant style so I went with them instead of single)
* Documentation and cleanup for
  * `clearPhoto()`
  * `_onionSkinChangeAmount()`
  * `canDisplayNews()`


BoatsAreRockable:
All looks good! :smile: 
I was wondering, why some are functions are prefixed by an underscore? Is it your intention for this to eventually be the case for all functions?

le717:
> I was wondering, why some are functions are prefixed by an underscore? Is it your intention for this to eventually be the case for all functions?

I am glad you asked! I am kinda dividing the functions into two categories; public and private. Private functions (the ones with underscores) could be considered more "low-level" or "core" functions that are not really (or ever) accessed by the HTML or called except in other "higher level" functions (ones without underscores). A good example would be `_deleteFile()`. It is a  "core" function, performing the actual file deletion while the functions that call it are

BoatsAreRockable:
Ah thanks a lot for explaining :smiley:

PR #20, opened 2015-10-16
Implement notification system

### LINKED ISSUE #19: User notification system
There are a number of error and success messages that ought to be displayed to the user. However, an `alert()` box creates a bad UX. What we need is a simple, elegant, unobtrusive system to display the messages.

We have two options:
1. The HTML5 [Notifications](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API) [API](http://www.paulund.co.uk/html5-notifications). It is simple and works well, but it looks different everywhere, can't be styled, and you cannot denote what type of message (success, general, error) is being displayed.
2. Roll our own, which is not that hard at all. There could be a div hidden at the top of the screen. When triggered, it will appear for a short bit to display the message and go away automatically. We can have 3 APIs: success, general, and warning. Each one can apply styling specific to the error type (success green, general blue, error red).

I am leaning more towards rolling our own for customization purposes, but I'll leave it up to you which should be used.

le717:
Fixes #19 and adds a few calls to the new system in the process. You will probably want to pull these changes locally and try them out before you merge, just to see how you like it. :wink:

I think this is the first real feature I've contributed. Everything else has been taking what already existed and improving it.

P.S. It may be time to revise the roadmap...

### DIFF
--- a/app/main.css
+++ b/app/main.css
@@ -396,7 +396,7 @@ button:active, a.button:active{
     z-index: 2;
 }
 #collapsedSidebar.shrink {
-    width: 50px;
+    width: 50px; 
 }
 
 #sidebar{
--- a/app/main.css
+++ b/app/main.css
@@ -396,7 +396,7 @@ button:active, a.button:active{
     z-index: 2;
 }
 #collapsedSidebar.shrink {
-    width: 50px; 
+    width: 50px;
 }
 
 #sidebar{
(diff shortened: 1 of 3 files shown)


app/main.css
```diff
@@ -557,3 +557,29 @@ input[type="range"]:hover{
 
 /* Frame rate input */
 #input-fr-change { margin-top: 0.3em; }
+
+/* ========== NOTIFICATION BAR ============== */
+.notification {
+  width: 60vw;
+  min-height: 3em;
+  position: absolute;
+  left: 18vw;
```
le717: I tried to make it roughly center. I didn't do an exact measurement or anything, so this may need tweaking.

BoatsAreRockable:
Yes the roadmap is becoming increasingly outdated! Perhaps we should have a proper discussion regarding the way forward once you've finished implementing your major cleanup and rewriting with #22

le717:
> once you've finished implementing your major cleanup and rewriting with #22

Believe me, I am not trying to take over the project and I probably won't be around forever. The thing is you have good plans and code that will support it, but it's kinda buried and tangled up. What I am doing is, as an experienced developer, taking the code and fixing it up so it may better support your plans (both current and future) and have a firm foundation to build on. :)

PR #21, opened 2015-10-16
Remove splash screen

### LINKED ISSUE #14: Remove first splash screen
When the program is opened, a splash screen is seen for a second, not enough time to read. Then, it opens up to the project manager box, which says the same thing as the splash screen (the name and the creator), along with other stuff. I think the first splash screen should be removed in order to save time, and because it really isn't needed, seeing as all of it's information is already seen on the next screen.

le717:
Fix #14. I went ahead and removed it as it was getting in my way and I was commenting it out so I could test quicker.


BoatsAreRockable:
Was just about to do this myself! :smile:

le717:
> Was just about to do this myself! :smile:

Sorry about that! I was just in dev mode and decided to go ahead and rip it out!

PR #22, opened 2015-10-16
[DO NOT MERGE] Rewrite picture taking code

le717:
Well, not a total rewrite as in none of your original setup is present anymore, but enough of a rewrite that it affects the entire code base and is not yet complete (hence the do not merge title).

This a a huge reworking, cleaning up the functions, following best practices, making them a bit more general, and as a bonus, supports individual frame deletion :smile:. This conflicts with the other PRs and I'm sure some of your local code too, so once it is done a point will have to be made to merge it and ensure everything else still works.

I am putting it up already so you can see how it works ahead of time and so we can ask questions before it lands.

### TODO ###
- [ ] Deleting an individual frame makes the frame reel act weird
- [x]  Undoing the last frame does not display a blank image in the frame reel

Issue #4, opened 2015-10-11
Frame scrolling arrows do not stop at 1st and last frame


BoatsAreRockable:
Scrolling beyond the first or last frame results in a buggy ui. Also the slider below the captured frames and the delete icons do not currently function.

le717:
Just wanted to say that all the issues listed here will be fixed in my next PR.

PR #38, opened 2015-10-23
Break off changes from #22 - Part 3

le717:
Here it is

This PR includes
* The remaining changes from #22 (specifically, rewriting `takePicture()`)
* A complete rewrite of the frame reel, which brought about:
  * Individual frame deletion
  * Frame reel scrolling
  * Fixes for all issues listed in #4 

The frame reel rewrite is the new code here. While thinking how I could improve the frame reel and make it work with the revised capture code, I concluded that the wheel was reinvented. There were 5 hard-coded frame previews, each one's `src` updated with the newly captured frame. Scrolling was planned to be done by clicking the button and updating the individual `src`s, which was a very intensive, slow, and complicated process, not to mention compounded the ability to delete individual frames. A scroll bar was going be added by way of a an `input type="range"` which presented an interesting set of challenges (how does sliding it change the frame reel? How do you go about making sure the scroll amount increases/decreases with the number of captured frames?) All this, I concluded, was reinventing what the browser can already do.

So I ripped it all out and started from scratch. I decided the previews would be dynamically injected into the reel via a `<table>`. This immediately solved many issues. Frames could be added/removed at will without having to worry about rebuilding the whole thing or holes appearing. A "No captured frames available." message would be displayed when (obviously) nothing was captured. This also meant in the future the reel can be dynamically built when frames are imported.

I then wrapped the table in a set-width container and set it to show a scroll bar when that value was exceeded. This would let the browser manage the scrolling and give the user more fine-grained scrolling control.

Finally, I made the scrolling arrows do their thing by adjusting the `scrollLeft` value of the container, again letting the browser manage everything.

I did have to do some minor markup changes to make the reel have the correct height at all times (the majority of the HTML changes are simple re-indentation) because `float`-based layouts are a pain and a half and the whole app should probably be reworked to not use them so much (Flexbox, what I used for the frame reel and the cause of the HTML changes, would be an excellent base. I have not followed it myself (I probably will hopefully soon), but I hear http://flexbox.io is a good place to learn all the ins and outs of Flexbox).

There you have it! :smile:


le717:
Oh snap, I just realized I forgot to restore the onion skin overlay. I will fix it, but it is still safe to merge this as-is. I don't want to keep these changes held up too much longer.

BoatsAreRockable:
Thank you very much, the frame reel was certainly the area I was previously most unhappy with. I'll have a look into Flexbox.

BoatsAreRockable:
I tested this out and it works fantastically :smile:

le717:
I think it's time to revise the roadmap so there is a clear idea of what features are desired and when. :)

PR #43, opened 2015-10-23
Merge play/pause buttons

### LINKED ISSUE #34: Merge the play and pause buttons
The Play Animation and Pause Animation buttons are separated. They should be combined. When you click play, it will turn into the pause button, so if you click it again, it will turn back into the play button. This helps for easy pausing when playing back an animation, because you don't have to move your mouse.

le717:
Fixes #34. The playback code will be the next thing I dive into.


BoatsAreRockable:
Oh have fun with that - you are most likely discovering some very odd functionality in terms of how looping works...

PR #44, opened 2015-10-23
Restore news feed

le717:
Pretty self-explanatory.

Screenshot:
news


BoatsAreRockable:
Merged. I think it would be good for the news feed element to have a scroll bar with a custom background colour so its blends in a bit better.

PR #45, opened 2015-10-24
Rioforce's Interface Revamp

rioforce:
In this pull request, I accomplished the following things:

* Organized entire HTML to be in a logical order, complete with comments
* Rewrote entire CSS to make a cleaner and more professional theme
* Made a dark theme (light theme will come when we add theme support)
* Made sidebar sticky
* Program is now responsive and works at multiple screen sizes
* Changed program icons (except for ones in menu bar) to use FontAwesome
* Reworked a bit of Javascript to make the interface unified

Unfortunately, there are plenty of bugs with the new interface, such as:

* Frames in frame reel don't preview
* Live View button in frame reel needs to be on right of frames
* Live View button is the wrong size
* Camera resolution is stuck at 480x360
* The frames are being upscaled to 640x480 to work with Onion Skinning
* When no frame is being previewed, the video and preview section shrinks (fix by adding placeholder div to hold the place of a frame)
* FontAwesome font won't work in the menu bar

Things that still need to be done:

* Revamp entire launcher window
* [Custom scrollbars](https://css-tricks.com/custom-scrollbars-in-webkit/)
* Notification styling
* Alert boxes need to be styled or converted to 
* All the issues filled on GitHub :wink:

There may be stuff that I forgot to mention. I'm eager to hear your opinion of it. Whether or not you actually merge it is up to you of course. ;)

### Screenshot ###
screenshot 255


BoatsAreRockable:
I think the new interface looks pretty sweet! :smiley: Although I would prefer a lighter theme as the default because the dark messes with the eye's perception of light and dark. I know that quite a few people would like the dark though, so perhaps theme support should become a priority.

I'm not too happy about how switching between capture and playback modes works though, it isn't very initiative. I'd prefer a written cue of which mode the user is in somewhere in the interface. The button's current location in the frame reel seems a bit weird to me.

A few other minor things:
* The delete last frame button should be the same size as the play and pause buttons
* There's a bug with the pause symbol not changing back to a play symbol
* Hover and click effects need to be re-added to the buttons

Most of the other bugs you mentioned were in the old interface anyway. Obviously this is quite a big pull release so I'll have a read through it all before I consider merging it and I will create a release of the code that is currently in the main repository. 

Overall I do like this redesign but I think it would be best to iron out the new bugs this has created before merging it. Although the biggest question is over whether a light or dark theme just be the default.

rioforce:
I'm glad you like the redesign, and yes, I do think it needs to be more stable before merging, I just thought I'd show it so in case you didn't like it, I wouldn't do all that work (and also I don't know how to fix a few of the things). If you create another branch of your repo, I can submit this to that branch and you can merge it in, allowing you to edit and the other collaborators to submit pull requests to that too, instead of replacing the master.

Personally, I think dark themes are better so the previewing image stands apart from the interface. It's all up to you though, but yeah, that's why I suggested theme support. ;)

As for the capture and playback switching, here's the logic behind it (and it doesn't work currently because the code is broken or something): When you click on one of the previous frames in the framereel, it will show it on the preview screen (a border will be added around the selected frame so you know what you are on. This also removes the need for a label saying what screen you are on.). When you click on the Live View icon (which might need a title or hover or something so you know what it is), it will show the live video as well as put a border around the icon so you know you're on it.

Basically, I tried designing the interface based on Dragonframe (but obviously less advanced and without the X-Sheet). It uses a similar frame-reel. Here's a pic:

dragonframe-example

I look forward to helping out with the interface more, this is a really great project! I wish I could help with JavaScript, but unfortunately, I have very limited knowledge of JS.

le717:
I like it. It is very clean, professional, and nice. Dark/light theme is a point of discussion, but rioforce has put all this together and it'd be kinda inconsiderate to dump it for a light theme. (Side point: theme support would be nice but there are still too many basic features missing for that to become a priority right now).

TBH, because this is a huge visual overhaul and introduces some new UX elements, I personally would rather iron out any new bugs specific to this pull request _then_ get back to working on the app itself. I really don't want to have to make changes then have to scrap/revise/redo them after this lands.

rioforce:
"it'd be kinda inconsiderate to dump it for a light theme"
Nah, it wouldn't be that at all. I don't mind at all, it's just a few colors to change. In fact, I just used three colors in the entire program. (Of course, a light theme would probably use more). It's all up to Boats.

BoatsAreRockable:
@le717 would it best to merge this straight away then?

BoatsAreRockable:
I've released the current source code as v0.5.0 and created a separate branch for it. As a result I shall proceed to merge this pull request. While this does add a couple of new bugs, they appear to be relatively fixable.

rioforce:
I was planning on re-submitting the pull-request to the separate branch instead of the master, but I suppose this works too, now that you've made a release. I'll file my list of knows issues so we can get to fixing them. ;)

PR #63, opened 2015-10-29
More fixes

rioforce:
Removing things out of the sidebar, adding them to logical places on the ui.

* The loop playback has been put beside the Play and Stop buttons.

* Frame Undo button has been restyled

* Onion skinning options have been moved below the framereel

* Framerate changer has been moved to below the framereel

Unfortunately, the dynamic positioning of the below-the-framereel items are a bit unstable, so I'll have to work on those later. Some other things like the Live View button, the frame reel, need to be styled still. I'd like to know what you think!

screenshot 256

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -44,22 +44,6 @@
                 </ul>
             </div>
 
-            <!--Playback Options-->
-            <div class="sidebar-item" id="playback-header">
-                <div class="sidebar-header">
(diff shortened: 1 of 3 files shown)


BoatsAreRockable:
Really like the loop button and the onion skinning is growing on me. I think the frame rate changing should be moved away from the onion skinning though because it looks like it is related somehow. Shall merge!

Issue #71, opened 2015-10-31
Onion Skinning is not very intuitive


rioforce:
While this is mainly my issue, since I redesigned the entire Onion Skinning slider, I thought I'd bring it up so we can discuss it. Basically, as it is, the onion skinning slider isn't very easy to understand. It is in the middle, and you can drag it up and down to change the onion skinning factor, but it only takes effect if the Onion Skinning button is enabled. The button and slider act awkwardly together, and it's not really easy to understand what each is at first glance.

So, I am thinking we need to find a way to get rid of the button and use the slider exclusively. One way to do this is to make the onion slider be at the bottom of the slider by default, and when you drag it up and down, it adjusts the opacity. This removes the need for a separate onion skinning button, but it makes the slider look off-balance with everything, since the circle isn't in the middle.

Another way to do it is to make the left and right portions of the slider do different things. It works like this in Dragonframe:
- Drag to the right: Onion skins last frame taken with live preview. If you select another frame, it doesn't show any live preview or onion skinning.
- Drag to the left: Onion skins last frame taken with live preview if on live mode. If you select another frame in the reel, it'll use that for onion skinning instead.

We don't have to do it like this, I'm just unsure what we need to do. Eliminating a need for a separate button would help a lot.
What do you guys think?

Robotic-Animations:
I agree that the button and slider is confusing, especially when using the program for the first time. I personally like the second proposal, as it seems like a more intuitive and versatile solution.

BoatsAreRockable:
Working towards a solution where the slider does different things based on if it is dragged left or right seems good. Although, as I mentioned in your pull request, the playback code must be sorted out first so this can be accommodated. 

We could partially implement this by removing the toggle button and having the last frame onion skin when the slider is dragged to the right. For now, moving the slider left of centre will do nothing.

BoatsAreRockable:
The keyframe system I'm currently working on in #116 could allow this to be implemented.

BoatsAreRockable:
If my keyframe system in #122 is implemented this feature could work so that sliding to the left shows the current keyframe as the onion skinning frame and sliding to the right shows the last captured frame.

le717:
If you think you can implement this once keyframes lands, then I'll go ahead and move this to 0.7.x.

le717:
@Robotic-Animations @rioforce You'll be happy to know this is currently in development at #122. 😃

charlielee:
[...]

PR #73, opened 2015-10-31
Compressed UI

rioforce:
In this revision, I shrunk down the below-the-video options and framereel. This adds much more room to the video preview, which is vital to animating. I also added the grid overlay button and the delete frame button.

IMO, the delete frame button on each frame in the framereel should go away, because it seems too easy to accidentally click. The main delete button would work better, because ideally, when you delete a frame, it automatically previews the next frame, thus, if someone was deleting many frames in a row, they wouldn't need to move their mouse, just click delete over and over again.

As you can see, I'm still struggling with the onion skin button. I made an issue for that (#71).

There are still issues with it, as usual, but it's fairly solid.

screenshot 257

What do you think?


BoatsAreRockable:
Perhaps the onion skinning could be made more initiative by only having the slider bar appear when it is activated, this would make it clear that the toggle button was related to it. Although this probably won't work so well when other things are added to the bottom toolbar.

The method you suggest in #71 to make the onion skinning slider more Dragonframe like would certainly make it much more powerful, although it is probably too complicated to implement until the frames on the frame reel are clickable. Correct me if I'm wrong, but I don't think it is very often that frames other than the last one captured are used for onion skinning. 

I think the delete buttons on the frame reel should be removed too.  When the ability to click on individual frames for a larger preview is implemented it would be great for the delete icon to show up in preview mode. Perhaps it could swap positions with the undo button so that the undo button only shows in capture mode and the delete button in playback mode.

Also I don't know about the capture button being yellow. It might grow on me but I think it would be fine if it was white.

I need to catch up on some sleep now, so I shall check out your branch and most likely merge in the morning.

BoatsAreRockable:
I shall merge this. I've suggested what should be done in the short term in regards to onion skinning in #71

PR #79, opened 2015-11-09
Revised Windows installer

le717:
* Removed standalone x64 installer by merging both into one installer (Inno will install the appropriate one automatically)
* Changed the Web2Exe output path to be more simple and cross-platform (and computer!) compatible
* Various installer fixes, including stronger compression and use of relative paths
* We now require Unicode Inno Setup to compile (which you should be using anyway...)

This is where I advocate dropping the x64 build because we're never actually going to use > 4 GB of RAM, longer build times, blah blah blah. I think x64 Chromium does have some JS speed improvements but I'm not sure if it is too much to count.


BoatsAreRockable:
Your arguments for dropping the x64 build seem pretty sound, I agree it should go. Perhaps in a while when the program is in a near complete state we can do some testing to see if there is any noticeable performance benefits.

Issue #83, opened 2015-11-15
Remove individual frame delete buttons


le717:
Spun off #73.

> The delete frame button on each frame in the framereel should go away, because it seems too easy to accidentally click. The main delete button would work better, because ideally, when you delete a frame, it automatically previews the next frame, thus, if someone was deleting many frames in a row, they wouldn't need to move their mouse, just click delete over and over again.

BoatsAreRockable:
One thought I had with this feature is that it if would be nice if the undo button only showed in capture mode and it was replaced by the delete button in playback mode.

le717:
I like that idea. Will implement.

le717:
Remove this from any milestone for now. The individual buttons currently works just fine for now, and I'd like to go ahead and submit this camera-related code. We will fix this, just not right now.

le717:
"Fixed" (more like nullified) by #124. Closing.

Issue #48, opened 2015-10-26
Live View button issues


rioforce:
* [ ] The live view button should not toggle between live and preview. When a frame on the framereel is selected, it will go to the preview frame, but if the live view button is selected, it will go to the live view.

* [ ] When the user is on live view mode, the live view icon needs to have a border around it, similar to that of when a frame is selected on the framereel.

* [x] The live view button is currently on the left side of the framereel, when it needs to be on the right, just like a frame on the framereel, it scrolls.

* [x] The live view button is also the wrong size, for some reason. 

* [x] The Live View button also needs a hover-title so people know what the button is.

Of course, this depends on if you keep the live view button the way it is. You mentioned you didn't think it was very good, but maybe it just needs a bit of help? We can discuss this. ;)

###### *Edited by BoatsAreRockable to make the bullet points into check boxes*

BoatsAreRockable:
I think it will be alright once it is moved to the far right of the frame reel and has a clear selected effect. I've added the current mode to the status bar to make it obvious which mode the user is in.

le717:
I have fixed a few of theses myself, will be in my next PR.

le717:
The remaining two items will be fixed once #90 is merged.

PR #90, opened 2015-12-04
Preview selected frame

### LINKED ISSUE #47: Click frames in framereel to show frame preview
When one clicks a frame in the framereel, it needs to show up in the preview box. It also needs to have a little border around the icon in the framereel so the user knows what frame they currently have selected.

In the StatusBar, the "Current frame" also needs to reflect the frame that is currently selected.

le717:
Fixes #47 (including playback) and the remaining items in #48. Allows #83 to be implemented.

### Notes ###
* When playing from a frame, if loop is enabled it will start back from the beginning.


le717:
Alright, bug is fixed. Ready for review. :)

app/js/main.js
```diff
@@ -254,19 +250,42 @@ function startup() {
...
+        if (document.querySelector(".frame-reel-preview.selected")) {
+            document.querySelector(".frame-reel-preview.selected").classList.remove("selected");
+            videoStop();
+        }
+
+        switchMode("capture");
+    });
+
+    // Preview a captured frame
+    document.querySelector("#reel-captured-imgs").addEventListener("click", function(e) {
```
BoatsAreRockable: Why is a querySelector used here rather than the `frameReelRow` variable?

app/js/main.js
```diff
@@ -197,17 +197,13 @@ function startup() {
     });
 
 
-/*==========================================================
-=============== LISTENERS ==================================
-===============================================================*/
-
-
+    /* ======= Listeners ======= */
```
BoatsAreRockable: I think I went a bit over the top with the header that I was using here before!

app/js/main.js
```diff
@@ -254,19 +250,42 @@ function startup() {
...
     });
 
     // Toggle capture and playback windows
     btnLiveView.addEventListener("click", function() {
-        if (winMode === "capture") {
-            winMode = "playback";
-        } else if (winMode === "playback") {
-            winMode = "capture";
+        // Switch from frame preview back to live view
+        if (document.querySelector(".frame-reel-preview.selected")) {
```
BoatsAreRockable: Would it be better for `.frame-reel-preview.selected` to be a new variable or to keep it like this?

BoatsAreRockable:
@le717 this looks good! :smile: I've added a couple of notes related to code style.

In terms of functionality you mention `When playing from a frame, if loop is enabled it will start back from the beginning.`. I'm not sure if this is a bug or intentional behaviour but I feel it is a bit unintuitive and should be changed.

I've found a similar issue when looping isn't enabled after playing from a frame and playback stopping at the last frame. If "play" is pressed again at this point playback starts from the first frame in the frame reel rather than the previously selected frame.

Another improvement I'd suggest is that when "stop" is pressed the white frame outline should move to the last frame in the frame reel.

Also on a side note when playback is in progress there should be some sort of visual cue in the timeline, such as a vertical line that moves along with each frame displayed. Although I appreciate this point is beyond the scope of this particular pull request so don't worry about it for now!

app/js/main.js
```diff
@@ -54,7 +54,7 @@ var width  = 640,
     // Status bar
     statusBarCurMode   = document.querySelector("#currentMode span"),
     statusBarCurFrame  = document.querySelector("#currentFrame span"),
-    statusBarFrameNum  = document.querySelector("#noOfFrames"),
+    statusBarFrameNum  = document.querySelector("#num-of-frames span"),
```
BoatsAreRockable: `#num-of-frames span` is an id that is in lower case with hyphens and doesn't match the the other ids used in the status bar. Would you recommend the the other ids in the status bar are changed to match this format?

app/js/main.js
le717: Because I did not know that variable existed and I didn't Ctrl+f the selector. Will fix.

app/js/main.js
le717: Ehhh, possibly.

app/js/main.js
le717: Yes, but it does not have happen now unless you want me to do it.

BoatsAreRockable: Yeah no need to do this for now

le717:
> Another improvement I'd suggest is that when "stop" is pressed the white frame outline should move to the last frame in the frame reel.

I tried this out. It was... _technically interesting_. However, I not sure it would be appropriate. First off, the outline does not move to each active frame during playback (your suggestion), so having it suddenly appear at stop would be odd. It also introduced a number of new cases where it'd need to be removed. I'll keep playing with it.

> In terms of functionality you mention "When playing from a frame, if loop is enabled it will start back from the beginning.." I'm not sure if this is a bug or intentional behaviour but I feel it is a bit unintuitive and should be changed.

I've found a similar issue when looping isn't enabled after playing from a frame and playback stopping at the last frame. If "play" is pressed again at this point playback starts from the first frame in the frame reel rather than the previously selected frame.

This was by design. There is no "previously played" state (and I don't think it'd be as simple as setting a true/false either. The frame preview is just that- a preview. It is only meant to temporarily alter the playback. If it were persistent, how would you reset it back to frame 1? A button in the UI? Maybe but could be confusing. Scroll back to frame 1, click it, then scroll back to the end and click the live view button? That is not acceptable. That's why I made it restart from the beginning.

That being said, I can still look into it if you would like.

BoatsAreRockable:
I see what your saying. There isn't an issue with what you've implemented at all, my previous suggestion would in fact make things far more confusing. 

What I'm really trying to work out is how to handle repeated playback not from the beginning. I think I've figured out how it could work based on the notation program Sibelius. I'll explain this in more detail tomorrow.

le717:
Loop region playback would be something to explore but it is a tricky thing to handle (and should not be a current focus).

BTW, this PR is done. The sooner you merge it, the sooner I can remove all those pesky delete buttons. ;)

BoatsAreRockable:
I've rechecked out your branch and I'm happy with how it works now and will therefore merge it. 

I'd like to take back what I said a couple of days about the playback reseting being "uninitiative". After thinking about it today I'm not sure why I saw a problem with it!

le717:
Hey, no need to take it back. it wasn't offensive or anything. It's called discussion and is a part of review. You brought up a path I had considered but did not go with. Since it's my PR, it was my burden of proof to explain how the implemented way was, even for the time being, better. It's all good, don't worry about it. :)

PR #93, opened 2015-12-09
(Even more) various fixes

le717:
Fixed these while working on #83. I would have waited to submit these but they are such obvious bugs that they should have already been fixed (+ one or two improvements and one important fix that was creating bugs).

* Make the Grid Overlay button use `notifyInfo()` instead `alert()`
* Update the current frame count in status bar when taking a new frame
* Set last taken frame (if possible) in playback window when deleting a
frame
* Automatically switch to capture mode if all frames are deleted
* Update frame preview selection when undoing a frame
* Fix a really bad listeners error in `confirmSet()` (`!important`)

I plan on writing a [blog post](https://triangle717.wordpress.com/) about the last one, but for a quick tl;dr here: the event listeners were getting adding to the buttons each time `confirmSet()` was called. If it was called three times, that means they received three listeners, meaning the callback would be run (you guessed it) three times. While `addEventListener` has safeguards to prevent identical, duplicate listeners, the way the function was written meant they did not apply. This patch makes sure the listeners are applied _once_ each the `confirmSet()` is called. As I said, I will detail more in an (eventual) post.

It's kinda my fault for letting this happen. In the back of my mind I knew this was happening but I didn't clearly acknowledge it. It's OK though. It is fixed now. :)

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -154,7 +154,7 @@
         <table id="control-panel-table">
             <tr>
                 <td id="onion-toggle-td">
-                    <div id="btn-grid-toggle"><a href="#" onclick="alert('Coming soon!');"><i class="fa fa-th" title="Grid Overlay"></i></a></div>
+                    <div id="btn-grid-toggle"><i class="fa fa-th" title="Grid Overlay"></i></div>
                     <div id="btn-onion-skin-toggle"><i class="fa fa-adjust fa-flip-horizontal" title="Enable Onion Skin"></i></div>
                 </td>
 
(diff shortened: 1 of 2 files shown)


BoatsAreRockable:
This could explain the previous issue I had with trying to make the confirm prompts focus on the OK button. I'm interested to see your blog post on this so give me a shout when it's done.

le717:
[Blog post is up.](https://triangle717.wordpress.com/2015/12/14/js-avoid-duplicate-listeners/)

PR #94, opened 2015-12-12
Automatically create the output folder if needed

### LINKED ISSUE #27: Automatically create the output folder if needed
If an output folder is set but does not actually exist, the path should be created and the user notified (use `notifyInfo()`) so frames can be saved/deleted.

BoatsAreRockable:
This fixes #27. If the selected export directory is deleted it will be recreated.

### Notes
* Currently only works if a single level is removed. ie if the output directory was at `C:/folder/subfolder` and `folder` was removed, `folder/subfolder` would not be recreated. Implementing a fix for this behaviour appears to be rather difficult in Node.js. Advice welcomed!
* The ability to add arguments to the callback has not been added.


le717:
Interesting implementation but not the best way to do it.

We should not use the callback pattern here. It actually complicates it more than it should be. Basically, if it has to create it, do so, leave a message in the console, and `return` to exit the function. If not, then immediately `return`. That's literally all the function had to do.

Once you have that, I'll look over the code again. See if you can find the most ideal place to call the function (is it in `startup()` or as part another function?). When the code is revised I'll also take a look for the ideal place and we'll compare findings. :wink:

le717:
> Currently only works if a single level is removed. ie if the output directory was at C:/folder/subfolder and folder was removed, folder/subfolder would not be recreated. Implementing a fix for this behaviour appears to be rather difficult in Node.js. Advice welcomed!

I looked into this and this repo to be the way to fix it: https://github.com/substack/node-mkdirp
Open a command prompt and `cd ` into `app`. Then run `npm install mkdirp --save`. Edit the .gitignore to block `node_modules`, and require the script per the directions in `app.js`. Tada! Now when you go to release Boats Animator you will need to package the node_modules folder for all to work, but it will all be fine. :+1: 

You got it! :D

BoatsAreRockable:
I'm working out where to call the function. I'm guessing using it in `takePicture()` every single time a frame is captured quite inefficient. `startup()` is probably the best place to call it at the moment however I see 2 issues with this:
* If the user has intentionally removed / renamed the default frame directory while BA has been closed it would be tedious for them to have to delete it again since it would be recreated upon opening the animator before they had a chance to change it.
* If the output directory is moved mid way through capture an error won't be detected and the remainder of the frames captured won't be exported.

However, I envisage the frame export settings eventually being in a separate preferences window that appears when creating a new project and can be accessed again later via the menu bar. This would mean that the user could change the export directory before the `startup()` function runs, therefore remedying the first the issue I've mentioned.

In order to fix the second issue the function should be called again when the program is going to be closed. If an error is detected as well as recreating the directory the captured frames should be exported again.

BoatsAreRockable:
@le717 I've finished adjusting this for now. Additionally I've added a confirm prompt when the animator is closed. The only thing I'm currently unhappy with is that if the export directory is deleted midway through animation, while the directory itself is recreated, the last frame captured and all frames prior to it are not re-exported.

app/js/main.js
```diff
@@ -658,7 +675,31 @@ function _getDefaultDirectory() {
 }
 
 /**
-* COnverting frames to png
+ * Check the save directory still exists and recreate it if deleted.
+ */
+function checkDirectoryExists() {
+    "use strict";
+    // Check if save directory exists.
+    fs.stat(frameExportDirectory, function (err, stats) {
```
le717: `mkdirp` calls `fs.stat` and ensures it does not exist before creating, so doing it here creates unneeded overhead.

app/js/main.js
```diff
@@ -658,7 +675,31 @@ function _getDefaultDirectory() {
...
+        if (err) {
+            // Tell the console that the directory doesn't exist.
+            console.log(`${frameExportDirectory} does not exist`);
+            // Attempt to recreate the missing directory.
+            mkdirp(frameExportDirectory, function (err) {
```
le717: This block here is pretty much the only code you need in the whole function.

app/js/main.js
```diff
@@ -658,7 +675,31 @@ function _getDefaultDirectory() {
 }
 
 /**
-* COnverting frames to png
+ * Check the save directory still exists and recreate it if deleted.
```
le717: Reword to:
"Create the default save directory if needed."

Simple and to the point.

app/js/main.js
```diff
@@ -658,7 +675,31 @@ function _getDefaultDirectory() {
...
+ */
+function checkDirectoryExists() {
```
le717: I almost want to rename this to `createSaveDirectory()` (generic name because we will reuse it later in projects).

app/js/main.js
```diff
@@ -111,10 +114,24 @@ function openIndex() {
...
 }
 
+/**
+ * Confirm prompt when animator is closed.
+ */
+win.on("close", function() {
+    confirmSet(closeAnimator, "", "Are you sure you to exit Boats Animator?");
+});
+
+function closeAnimator() {
```
le717: We don't call `closeAnimator()` anywhere else, so merge it into the `confirmSet()` call.

app/js/main.js
```diff
@@ -111,10 +114,24 @@ function openIndex() {
     });
 }
 
+/**
+ * Confirm prompt when animator is closed.
+ */
+win.on("close", function() {
+    confirmSet(closeAnimator, "", "Are you sure you to exit Boats Animator?");
```
le717: Rewrite as so:
`confirmSet(win.close, true, "Are you sure you to exit Boats Animator?");`

le717:
> If the output directory is moved mid way through capture an error won't be detected and the remainder of the frames captured won't be exported.

> The only thing I'm currently unhappy with is that if the export directory is deleted midway through animation, while the directory itself is recreated, the last frame captured and all frames prior to it are not re-exported.

While a valid concern, the thing I have learned in my nearly 3 years of programming (a lot of it has been related to file processes) is _this never intentionally happens_. In fact, it rarely ever happens, intentionally or not.

My advise: it is not worth the time or trouble to manage. In some applications it is proper to handle the case, but not in the majority of apps (including BA). :smile:

app/js/main.js
```diff
@@ -725,9 +766,11 @@ function _writeFile(file, data) {
     "use strict";
      fs.writeFile(file, data, function(err) {
         if (err) {
-            throw err;
+            console.error(`Failed to save ${file} to disk`);
+            checkDirectoryExists();
```
le717: This line goes back to my comment [here](https://github.com/BoatsAreRockable/animator/pull/94#issuecomment-165914811) but I think this will be fine to keep. I am concerned that you dropped the `err` object. Maybe you can `console.error(err);` that? It can provide vital information.

le717:
>  I'm guessing using it in takePicture() every single time a frame is captured quite inefficient.

You guess correctly.

> If the user has intentionally removed / renamed the default frame directory while BA has been closed it would be tedious for them to have to delete it again since it would be recreated upon opening the animator before they had a chance to change it.

> However, I envisage the frame export settings eventually being in a separate preferences window that appears when creating a new project and can be accessed again later via the menu bar. This would mean that the user could change the export directory before the startup() function runs, therefore remedying the first issue I've mentioned.

First, you must understand that projects is a drastic change that will change most everything we currently do concerning the save path. A key change is there will be none of this auto-creation if missing code you are adding.

In the current, project-less state we need the functionality in this PR because there is no other way to save files. With projects, the folders will be created upon project creation- and that is it (with the exception of new takes). If the project is moved we do not attempt to find or recreate it, we respond in the universal, _expected_ behavior: tell the user we cannot find it and remove it from the recent projects list. The user then opens a project, navigating to the new project location, and we load it (and possibly recreate any missing folders, but that is an implementation detail). There will be no way within BA to move the project to a different directory (that takes some seriously crazy code and is out of app purpose) because they can do so in Explorer/Finder/whatever Linux uses.

I applaud you for thinking forward and planning for the future, but the future will be drastically different in this area. :)

Back to the here-and-now, I agree with your assessment of checking it in `startup()`. Per your first concern, yes it would be tedious for the user, but for the time being they'll need to change it in BA first _then_ change the folder. Yes, even this approach has issues, but the system is less than ideal, so we are having to put up with less than ideal behavior.

BoatsAreRockable:
@le717 Sorry for the delay - I'm currently away from home. I've pushed most of the further changes you suggested, however I was unable to merge `closeAnimator()` with the `confirmSet()`. Doing so caused pressing OK in the close dialogue to have no effect. I suppose this is down to the less than ideal implementation of confirmSet in the first place

le717:
> I was unable to merge closeAnimator() with the confirmSet(). Doing so caused pressing OK in the close dialogue to have no effect.

Meh. It's fine.

This looks good, but we to do one more thing. There are 20 commits in this PR! We don't need all the various revisions in the commit history. What we need to do is flatten them using the `git rebase` command.

Note: you normally do not rebase pushed commits in case someone locally checked out the branch, but I did not so we are good. :)

1.Open a Git shell to the git repo (press the tilde key in GitHub Desktop). Make sure you are on this branch.
2. Run the following command: `git rebase -i HEAD~20`. `20` because this branch has 20 commits we want to rebase.
3. Change the first commit action to `edit` and the rest to `fixup`.
4. Rewrite the first commit message to be more accurate
5. Once rebasing is done, run `git push -f`. This will force push the new, single commit and rewrite the branch history on the server.
6. Let me know when you are done and/or have any questions. :)

BoatsAreRockable:
@le717 I've combined the commits although there are now a few edits conflicts.

le717:
OK, that is weird. It looks like you did a `git merge` to pull in the latest changes from `master` then rebased the branch. If so, you should have rebased your branch _then_ merged and stopped. Thus PR would have two commits: the feature commit and a merge commit.

We are just going to take the easy route here. Backup your code from this branch, close this PR, make a _new_ branch off an up-to-date `master`, overwrite it with the changed files from here (reverting any changes that were accidentally made from this), then make a new PR and ping me when it is up. :)

PR #100, opened 2015-12-20
Fix very tiny bug with border on capture window

le717:
When the `active` class is add/removed to/from `#captureWindow`, the element shrinks/grows by 2px. This PR fixes that. :)


BoatsAreRockable:
Ah yes this was something I noticed! :)

PR #102, opened 2016-01-11
Automatically create output folder if required (2)

BoatsAreRockable:
@le717 I think this is finally sorted now. It's amazing how much trouble such a small amount of code has caused me!

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -675,12 +675,13 @@ function _getDefaultDirectory() {
 }
 
 /**
- * Create the default save directory if needed
+ * Create the default save directory if needed.
  */
 function createSaveDirectory() {
     "use strict";
     mkdirp(frameExportDirectory, function (err) {
         if (err) {
+            console.error(err);
             console.error(`Failed to create directory at ${frameExportDirectory}`);
             notifyError(`${frameExportDirectory} does not exist. Failed to create a directory at this location.`);
         } else {
@@ -758,7 +759,7 @@ function _writeFile(file, data) {
     "use strict";
      fs.writeFile(file, data, function(err) {
         if (err) {
-            console.error(`Failed to save ${file} to disk`);
+            console.error(err);
             createSaveDirectory();
         } else {
             console.log(`Successfully saved ${file} to disk`);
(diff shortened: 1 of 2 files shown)


le717:
I'll take a look ASAP, hopefully after work today.

app/js/main.js
```diff
@@ -658,7 +675,23 @@ function _getDefaultDirectory() {
 }
 
 /**
-* COnverting frames to png
+ * Create the default save directory if needed
```
le717: Nit (for future reference): all function descriptions should end with a period.

app/js/main.js
```diff
@@ -658,7 +675,23 @@ function _getDefaultDirectory() {
...
+ */
+function createSaveDirectory() {
+    "use strict";
+    mkdirp(frameExportDirectory, function (err) {
+        if (err) {
+            console.error(`Failed to create directory at ${frameExportDirectory}`);
+            notifyError(`${frameExportDirectory} does not exist. Failed to create a directory at this location.`);
```
le717: Add `console.error(err);` before the call to `notifyInfo()` so we can know exactly what happened.

app/js/main.js
```diff
@@ -725,9 +758,11 @@ function _writeFile(file, data) {
     "use strict";
      fs.writeFile(file, data, function(err) {
         if (err) {
-            throw err;
+            console.error(`Failed to save ${file} to disk`);
```
le717: Again, add `console.error(err);` before calling `createSaveDirectory()` so we can get details.

app/js/main.js
```diff
@@ -725,9 +758,11 @@ function _writeFile(file, data) {
...
+            createSaveDirectory();
```
le717: I'm not so sure about this call. It's assuming the error was caused by a missing output folder, which is not always the case. We also talked about the unlikelyhood of the output folder being ripped out from under the app.

I'll leave it up to do. Do what you think is best.

le717:
Aside from my comments, looking good! :+1:

le717:
BTW, you should usually delete branches once the PR has been merged. In this case, you can delete `issue-27` and `issue-27-new`. ;)

BoatsAreRockable:
Alright, I will do that now :)

PR #106, opened 2016-01-31
Fix unset save directory on launch

### LINKED ISSUE #104: App does not fully initialize on load if save path is null
Repo steps:
* Clear the save path
* Close app
* Reopen app, start new project
* Notice app does not fully load
* Open dev tools, see error message

error

Introduced in #102.

le717:
Fix #104, introduced in #102. If a save directory is not set on launch, immediately prompt to set it.

Also revise related messages to have consistent wording.

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -72,10 +72,6 @@
                     <li>
                         <i class="fa fa-sort-asc fa-rotate-90 sidebar-link-dot"></i>
                         <span class="sidebar-opt" id="btn-dir-change">Change directory</span>
-                    </li>
-                    <li>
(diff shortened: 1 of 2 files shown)


le717:
OK, I've just pushed a totally different commit. I was wondering if my fix was the best way to do it, and while looking at it I got frustrated by the code and UX and reworked it. Now, when a user picks a path, it is automatically saved _without_ clicking a separate link. This is actually expected behavior.

(I will be so happy when projects are eventually reached...)

BoatsAreRockable:
Automatically saving the path does seem more intuitive.

PR #107, opened 2016-02-01
New `file` module

le717:
This moves and generalizes all low-level `fs` functions in a new `file` module. In this way, all file system operations can be consistent, reusable, expandable, and all in one place. Modularity is a good thing, and this is a good area to do that.

I would like this to make it into the last 0.6.x release if possible. Even if not, once the other two PRs are merged I will need to update it. I suspect there will be a few merge conflicts...


le717:
Aha, as I suspected: a merge conflict! I will fix this right up.

le717:
Conflict fixed, ready for merging.

Issue #112, opened 2016-02-03
Improving playback mode


BoatsAreRockable:
There are a few things that I'd like to add to improve the frame reel area and playback. I may end up splitting this into multiple issues.
- [x] Additional playback controls #141
- [ ] Play / loop selection #116
- [x] Indication of current frame being played #115
- [x] Framereel scrollbar #163
## Additional playback controls

_Split out as #141_
As well as eventual keyboard shortcuts there should be buttons in the interface for previewing the previous and next frame relative to the currently selected one as well as buttons for jumping to the first and last frame captured. (basically buttons for `|<`, `<`, `>` and `>|`)

**Although:** I'm debating whether these require buttons as well keyboard shortcuts. I don't want the controls area getting too cluttered.
## Play / loop selection of captured frames

_Split out as #116_
After selecting a frame 3 more buttons (with keyboard shortcuts) will become available in the playback controls: (These will have icons) "Begin selection", "End selection" and "Unselect". 
- **Begin selection** - pressing this button after selecting a frame in the framereel will add a _blue?_ vertical line before it and means that after playback stops pressing play will start playback again from this point
  - If only "begin selection" is used then playback will run from the selected frame to the last frame.
- **End selection** - this button will be greyed out if "begin selection" hasn't been used yet. Selecting it will add a blue vertical line after it and mean that playback will run from the frame selected with "begin selection" to here.
  - If the frame selected with this tool is before the frame chosen with "begin selection" then playback will instead run from this frame to the "begin selection" frame.
- "Unselect" - as its name suggests, pressing this button removes both the start and end selection points
- The two selection points should eventually be draggable along the frame reel.
## Indication of current frame being played in the frame reel

_Split out as #115_
There should be a "position" arrow in the frame reel to indicate which frame is currently being played. The frame reel should scroll with this during playback. It should disappear after a frame is selected or when switching to capture mode.

I've currently been testing the positioner out and will create a pull request soon: 

<img width="904" alt="ba frameposition" src="https://cloud.githubusercontent.com/assets/8095888/12789326/3f8093c4-ca95-11e5-8e50-ebe114905bfe.png">
## Framereel scrollbar

_Split out as #163_
I implemented this is in a rush without properly consulting the other contributors to this project (This was also implemented before I knew how to use branches and PRs...), as a result I'll happily say it looks rather ugly!

It should be restyled with a darker background colour and probably be squarer.

Lots of ideas here, feedback is much appreciated!

[...]

PR #113, opened 2016-02-03
Indication on frame reel of current frame being played

BoatsAreRockable:
This is part of #112 (and now #115). When playback occurs a red line moves along the frame reel and is displayed after the frame in the reel corresponding to the frame currently being previewed (this is easier it explain visually than in words!).

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -232,7 +232,7 @@ function startup() {
     // Stop the preview
     btnStop.addEventListener("click", function() {
         if (curFrame > 0) {
-            removeFrameReelSelection();
+            _removeFrameReelSelection();
             videoStop();
         }
     });
@@ -264,7 +264,7 @@ function startup() {
     // Switch from frame preview back to live view
     btnLiveView.addEventListener("click", function() {
         videoStop();
-        removeFrameReelSelection();
+        _removeFrameReelSelection();
         switchMode("capture");
     });
 
@@ -272,7 +272,7 @@ function startup() {
     frameReelRow.addEventListener("click", function(e) {
         if (e.target.className === "frame-reel-img") {
             // Remove previous selection
-            removeFrameReelSelection();
+            _removeFrameReelSelection();
 
             // Highlight the clicked image
             e.target.classList.add("selected");
@@ -318,7 +318,7 @@ function switchMode(newMode) {
  *
  * @return {Boolean} True if there was a highlight to remove, false otherwise.
  */
-function removeFrameReelSelection() {
+function _removeFrameReelSelection() {
     "use strict";
     var selectedFrame = document.querySelector(".frame-reel-img.selected");
     if (selectedFrame) {
@@ -327,6 +327,15 @@ function removeFrameReelSelection() {
         return true;
     }
     return false;
+}
+
+/**
+ * Add selected frame highlight to frame.
+ * @param {Number} id The image ID to highlight.
+ */
+function _addFrameReelSelection(id) {
+    "use strict";
+    document.querySelector(`.frame-reel-img#img-${id}`).classList.add("selected");
 }
 
 /**
@@ -347,7 +356,7 @@ function updateFrameReel(action, id) {
     // Add the newly captured frame
     if (action === "capture") {
         // Remove any frame selection
-        removeFrameReelSelection();
+        _removeFrameReelSelection();
 
         // Insert the new frame into the reel
         frameReelRow.insertAdjacentHTML("beforeend", `<td><div class="frame-reel-preview">
@@ -383,7 +392,7 @@ function updateFrameReel(action, id) {
 
         // Update frame preview selection
         if (curSelectedFrame) {
-            removeFrameReelSelection();
+            _removeFrameReelSelection();
             document.querySelector(`.frame-reel-img#img-${id - 1}`).classList.add("selected");
         }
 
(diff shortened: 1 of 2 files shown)


app/js/main.js
```diff
@@ -559,6 +563,21 @@ function _videoPlay() {
     removeFrameReelSelection();
     playback.setAttribute("src", capturedFramesRaw[curPlayFrame]);
     statusBarCurFrame.innerHTML = curPlayFrame + 1;
+
+    // Display playback position indicator after each frame
+    document.getElementById("img-" + (curPlayFrame + 1)).style.borderRight = "4px dotted red";
```
BoatsAreRockable: Is it worth making `.style.borderRight` a CSS class that is toggled?

le717:
Drive-by comment: use querySelector, not getElementBy*. Also, use template strings to construct the selector.

app/js/main.js
```diff
@@ -576,6 +595,15 @@ function _videoPlay() {
 }
 
 /**
+ * Clear the red playback position indicator.
+ */
+function _clearReelPositioner() {
+    for (i = 1; i < (curFrame + 1); i++) {
```
le717: `for (var i = 1;...` ;)

app/js/main.js
```diff
@@ -576,6 +595,15 @@ function _videoPlay() {
 }
 
 /**
+ * Clear the red playback position indicator.
```
le717: Remove "red" from this sentence.

app/js/main.js
le717: Yes, it is.

le717:
I honestly thought I already implemented this. Guess not.

Instead of the red line, why not reuse `.frame-reel-img.selected`? I actually intended for that to be used for selection and during playback. It's why I implemented it the way I did (`removeFrameReelSelection()` and all).

le717:
I knew I (mostly) implemented that already! 35a899dd3f492f86318d62bc0cb91bebe1728cb7

Grab the document.querySelector line and move it into a new function:

```js
function _addFrameReelSelection(id) {
    "use strict";
    document.querySelector(`.frame-reel-img#img-${id}`).classList.add("selected");
}
```

Then use this every time you need to add the highlight. (While you are at it you can prefix `removeFrameReelSelection` with that missing underscore.)

app/js/main.js
```diff
@@ -559,6 +563,21 @@ function _videoPlay() {
...
+
+    // Display playback position indicator after each frame
+    document.getElementById("img-" + (curPlayFrame + 1)).style.borderRight = "4px dotted red";
+    if (curPlayFrame > 0) {
+        document.getElementById("img-" + curPlayFrame).style.borderRight = "none";
+    }
+
+    if (curPlayFrame + 1 !== curFrame) {
+        // Scroll so currently played frame is in view
+        document.getElementById("img-" + (curPlayFrame + 1)).scrollIntoView();
```
le717: I do like this idea. +1

le717:
I didn't know you pushed any changes. You have to say so. GitHub doesn't send out emails for that. ;)

BoatsAreRockable:
@le717 Ah my bad!

le717:
I'll try to look at this this afternoon. ;)

le717:
Ack! I forgot about this! Sorry! I'll review it right now!

app/js/main.js
```diff
@@ -330,6 +330,15 @@ function removeFrameReelSelection() {
 }
 
 /**
+ * Add selected frame highlight to frame.
+ * @param {Number} id The image ID to highlight.
```
le717: Blank line right here above `@param` (see the JSDoc annotations in file.js for the format we'll transition to once I write down some things and get it to you).

app/js/main.js
```diff
@@ -556,9 +565,24 @@ function videoStop() {
...
     playback.setAttribute("src", capturedFramesRaw[curPlayFrame]);
     statusBarCurFrame.innerHTML = curPlayFrame + 1;
+
+    // Display playback position indicator after each frame
+    _addFrameReelSelection(curPlayFrame + 1);
+    
+    if (curPlayFrame === 0) {
+        // Scroll to start when playback begins
+        frameReelArea.scrollLeft = 0;
+    } else if (curPlayFrame + 1 !== curFrame) {
```
le717: Why do we keep having to deal with this `+ 1` mess? :(

le717:
I don't like how the frame reel scrolling is in `videoPlay()`. I mean, it could call a function to interact with the frame reel but it should not be a part of the fplayback code.. Refactor that code into a new function and name it so it is clear it deals with the frame reel and have `_videoPlay()` call it instead.

app/js/main.js
BoatsAreRockable: @le717 curFrame is not zero based because each time a frame is captured it is incremented this means if 20 frames have been captured curFrame = 20.

BoatsAreRockable:
@le717 Just pushed some more changes to this.

le717:
This looks better. I'll do one last look when I'm on my laptop.

le717:
BTW, looking at this right now.

app/js/main.js
```diff
@@ -594,6 +610,22 @@ function previewCapturedFrames() {
 }
 
 /**
+ * Scroll the framereel during playback
+ */
+function _frameReelScroll() {
+    if (curPlayFrame === 0) {
```
le717: This function is missing a `"use strict;" directive.

le717:
Alright. Aside from the missing strict directive, this looks fine. After fixing that, squash the commits (`git rebase` then `git push -f`, remember?) and you can immediately merge.

BoatsAreRockable:
@le717 I think I managed to rebase and merge it correctly this time!

le717:
Indeed it does!

Issue #117, opened 2016-02-18
Playback window slightly larger than capture window.


BoatsAreRockable:
The issue that #100 fixed in regards to the preview window changing sizes between modes seems to have reappeared. This also means that onion skinning does not line up perfectly.

PR #118, opened 2016-02-18
Fix #65 - onion skinning breaks if preview area is taller than it is wide and #117

### LINKED ISSUE #65: Onion skinning breaks if preview area is taller than it is wide
Imgur

This bug comes from the quite horrid current method of determining the height of the preview area.

BoatsAreRockable:
I've fixed the long term issue #65 and took the opportunity to remove some unnecessary "layer" elements. It took me a while to realise that the reason issue #65 didn't affect `#preview` was because WebKit assigns it the property `object-fit: contain;` by default - something I haven't come across before!

I'm also happy to say to say that the dodgy aspect ratio specific `calc()` widths are now gone.

Issue #117 is also fixed in this request with a transparent border being applied to the playback window.

### DIFF
--- a/app/css/animator.css
+++ b/app/css/animator.css
@@ -172,57 +172,54 @@ a {
 
 #captureWindow, #playbackWindow {
   height: calc(100vh - 200px);
-/*  height: 30em;*/
+  width: 100%;
   text-align: center;
   background-color: black;
+  position: relative;
+  border: 2px solid transparent;
 }
 
 #captureWindow.hidden,
 #playbackWindow.hidden { display: none; }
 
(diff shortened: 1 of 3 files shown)


app/css/animator.css
```diff
@@ -170,59 +170,56 @@ a {
...
-
-.windowTitleContainer {
-  text-align: center;
-  display: inline-block;
+  object-fit: contain;
 }
 
 #playback {
-    height: calc((100vh - 200px));
-    width: calc(((100vh - 200px) / 3) * 4);
```
BoatsAreRockable: Very pleased to see this go...

PR #119, opened 2016-02-20
Get the app version using built-in manifest object

le717:
I saw this [in the docs](
http://docs.nwjs.io/en/v0.13.0-beta6/References/App/#appmanifest) and instantly knew it would be better than parsing the manifest file ourselves.


le717:
Wow, that was fast. :P

PR #120, opened 2016-02-20
It's 2016

le717:
'Nuff said.


BoatsAreRockable:
Oops... This will eventually be gone from the status bar when an about window is created

PR #122, opened 2016-02-29
Keyframes 1 fixes #116

### LINKED ISSUE #116: Play / loop selection of captured frames
_Part of #112_

> After selecting a frame 3 more buttons (with keyboard shortcuts) will become available in the playback controls: (These will have icons) "Begin selection", "End selection" and "Unselect". 
> 
> * **Begin selection** - pressing this button after selecting a frame in the framereel will add a _blue?_ vertical line before it and means that after playback stops pressing play will start playback again from this point
>   * If only "begin selection" is used then playback will run from the selected frame to the last frame.
> 
> * **End selection** - this button will be greyed out if "begin selection" hasn't been used yet. Selecting it will add a blue vertical line after it and mean that playback will run from the frame selected with "begin selection" to here.
>   * If the frame selected with this tool is before the frame chosen with "begin selection" then playback will instead run from this frame to the "begin selection" frame.
> 
> * "Unselect" - as its name suggests, pressing this button removes both the start and end selection points
> 
> * The two selection points should eventually be draggable along the frame reel.

I have begun working on this.

BoatsAreRockable:
I've had a few attempts at working out how to do this but believe this implementation works quite well.

In this pull request:
* 2 new buttons are added to the playback controls: "begin selection" and "clear selection" (`btn-start-keyframe` and `btn-clear-keyframe`). These are disabled with their opacity reduced when not required.
* 2 new functions added: `addKeyframe()` and `removeKeyframe()`. These have been implemented in such a way to allow to allow additional functionality to be added to the keyframe system in the future. These functions essentially set/clear a new variable `curStartKeyframe` and add/remove visual cues on the framereel.
* Changes to the playback code have been made to see if a `curStartKeyframe` has been set.

My more detailed functionality plan:
[Play Selection Plan.docx](https://github.com/BoatsAreRockable/animator/files/150963/Play.Selection.Plan.docx)

In #116 I suggested an `end selection` button as well. However, this created a number of complications and made this feature a much more confusing UX.

@le717 I'd welcome your feedback on this. Thanks! :)

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -377,6 +377,7 @@ function _removeFrameReelSelection() {
 function _addFrameReelSelection(id) {
     "use strict";
     document.querySelector(`.frame-reel-img#img-${id}`).classList.add("selected");
+    curSelectedFrame = id;
 }
 
 /**
@@ -627,6 +628,10 @@ function videoPause() {
         btnPlayPause.children[0].classList.remove("fa-pause");
         btnPlayPause.children[0].classList.add("fa-play");
         console.info("Playback paused");
+
+        // Enable keyframe options
+        btnStartKeyframe.removeAttribute("disabled");
+        btnClearKeyframe.removeAttribute("disabled");
     }
 }
 
@@ -637,7 +642,12 @@ function videoStop() {
     "use strict";
     // Reset the player
     videoPause();
-    curPlayFrame = 0;
+    if (curStartKeyframe) {
+        // Reset playback
+        curPlayFrame = curStartKeyframe - 1;
+    } else {
+        curPlayFrame = 0;
+    }
     playback.setAttribute("src", capturedFramesRaw[curFrame - 1]);
 
     // Display newest frame number in status bar
@@ -670,10 +680,13 @@ function _videoPlay() {
             videoStop();
         } else {
             console.info("Playback looped");
+            if (curStartKeyframe) {
+                // Reset playback
+                curPlayFrame = curStartKeyframe - 1;
+            } else {
+                curPlayFrame = 0;
+            }
         }
-
-        // Reset playback
-        curPlayFrame = 0;
     }
 }
 
@@ -688,6 +701,10 @@ function previewCapturedFrames() {
     // Update the play/pause button
     btnPlayPause.children[0].classList.remove("fa-play");
     btnPlayPause.children[0].classList.add("fa-pause");
+
+    // Disable keyframe options
+    btnStartKeyframe.setAttribute("disabled", "");
+    btnClearKeyframe.setAttribute("disabled", "");
 
     // Begin playing the frames
     isPlaying = true;
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -49,6 +49,11 @@ var width  = 640,
     btnPlayPause  = document.querySelector("#btn-play-pause"),
     inputChangeFR = document.querySelector("#input-fr-change"),
 
+    // Keyframes
+    btnStartKeyframe  = document.querySelector("#btn-start-keyframe"),
+    btnClearKeyframe  = document.querySelector("#btn-clear-keyframe"),
+    curStartKeyframe  = null,
+
     // Audio
     captureAudio = new Audio("audio/camera.wav"),
     audioToggle  = document.querySelector("#audio-toggle"),
@@ -270,22 +275,53 @@ function startup() {
 
     // Preview a captured frame
     frameReelRow.addEventListener("click", function(e) {
-        if (e.target.className === "frame-reel-img") {
-            // Remove previous selection
-            _removeFrameReelSelection();
+        if (e.target.className.indexOf("frame-reel-img") > -1) {
+            if (e.target.className.indexOf("selected") === -1) {
+                // Remove previous selection
+                _removeFrameReelSelection();
 
-            // Highlight the clicked image
-            e.target.classList.add("selected");
-            if (winMode !== "playback") {
-                switchMode("playback");
+                // Highlight the clicked image
+                e.target.classList.add("selected");
+                if (winMode !== "playback") {
+                    switchMode("playback");
+                }
+
+                // Display the image and update all the necessary values
+                var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
+                playback.setAttribute("src", capturedFramesRaw[imageID - 1]);
+                curPlayFrame = imageID - 1;
+                curSelectedFrame = imageID;
+                statusBarCurFrame.innerHTML = imageID;
+
+            } else if (e.target.className === "frame-reel-img selected") {
+                    addKeyframe("start", curSelectedFrame);
+
+            } else if (e.target.className.indexOf("keyframe") > -1) {
+                // If the current start keyframe is double clicked it is removed
+                if (parseInt(e.target.id.slice(4)) === curStartKeyframe) {
+                    removeKeyframe("start");
+                }
             }
+        }
+    });
 
-            // Display the image and update all the necessary values
-            var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
-            playback.setAttribute("src", capturedFramesRaw[imageID - 1]);
-            curPlayFrame = imageID - 1;
-            curSelectedFrame = imageID;
-            statusBarCurFrame.innerHTML = imageID;
+    // Add a start keyframe
+    btnStartKeyframe.addEventListener("click", function(e) {
+        if (btnStartKeyframe.hasAttribute("disabled")) {
+            return false;
+        } else if (curSelectedFrame === curStartKeyframe) {
+            removeKeyframe("start");
+        } else {
+            addKeyframe("start", curSelectedFrame);
+        }
+    });
+
+    // Remove the start keyframe
+    btnClearKeyframe.addEventListener("click", function(e) {
+        if (!curStartKeyframe) {
+            return false;
+        } else {
+            removeKeyframe("start");
         }
     });
 }
@@ -302,12 +338,17 @@ function switchMode(newMode) {
         captureWindow.classList.remove("hidden");
         captureWindow.classList.add("active");
         btnLiveView.classList.add("selected");
+        btnStartKeyframe.setAttribute("disabled", "");
+        if (!curStartKeyframe) {
+            btnClearKeyframe.setAttribute("disabled", "");
+        }
 
     } else if (winMode === "playback") {
         playbackWindow.classList.remove("hidden");
         captureWindow.classList.add("hidden");
         captureWindow.classList.remove("active");
         btnLiveView.classList.remove("selected");
+        btnStartKeyframe.removeAttribute("disabled");
     }
     console.log(`Switched to: ${winMode}`);
     statusBarCurMode.innerHTML = winMode.charAt(0).toUpperCase() + winMode.slice(1);
@@ -339,7 +380,7 @@ function _addFrameReelSelection(id) {
 }
 
 /**
- * Update the frame reel display as needeed.
+ * Update the frame reel display as needed.
  *
  * @param {String} action Update the frame reel depending on the
  *                        action performed. Possible values are
@@ -405,6 +446,47 @@ function updateFrameReel(action, id) {
 }
 
 /**
+ * Set the selected frame as a keyframe
+ * @param {String} location The type of keyframe to be added.
+ *                          Currently accepted value is "start".
+ * @param {Integer} id Which frame should be set as a keyframe.
+ */
+function addKeyframe(location, id) {
+    "use strict";
+    if (location === "start") {
+        // Remove the previous start keyframe
+        if (curStartKeyframe) {
+            removeKeyframe("start");
+        }
+
+        btnClearKeyframe.removeAttribute("disabled");
+
+        // Add a new start keyframe
+        btnStartKeyframe.classList.add("active");
+        document.querySelector(`.frame-reel-img#img-${id}`).classList.add("keyframe");
+        curStartKeyframe = id;
+        console.info(`New startKeyframe: frame ${id}`);
+    }
+}
+
+/**
+ * Remove the selected keyframe.
+ * @param {String} location Which keyframe should be removed.
+ *                          Currently accepted value is "start".
+ */
+function removeKeyframe(location) {
+    "use strict";
+    if (location === "start") {
+        btnStartKeyframe.classList.remove("active");
+        document.querySelector(`.frame-reel-img#img-${curStartKeyframe}`).classList.remove("keyframe");
+        curStartKeyframe = null;
+        console.info(`Removed curStartKeyframe`);
+    }
+
+    btnClearKeyframe.setAttribute("disabled", "");
+}
+
+/**
  * Delete an individual frame.
  *
  * @param {Number} id The frame ID to delete.
@@ -416,6 +498,10 @@ function deleteFrame(id) {
             notifySuccess("File successfully deleted.");
         }
     });
+
+    if (id === curStartKeyframe) {
+        removeKeyframe("start");
+    }
 
     exportedFramesList.splice(id - 1, 1);
     capturedFramesRaw.splice(id - 1, 1);
--- a/app/animator.html
+++ b/app/animator.html
@@ -124,9 +124,11 @@
                             </div>
                     </td>
                     <td id="playback-controls">
-                            <div id="btn-loop" class="mediaControls" title="Loop Playback"><i class="fa fa-refresh"></i></div>
-                            <div id="btn-stop" class="mediaControls" title="Stop Playback"><i class="fa fa-stop"></i></div>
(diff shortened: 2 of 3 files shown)


le717:
Yea, you'll have to hold your horses on any work from me anytime soon. :(

https://twitter.com/_le717/status/704477544620417027

BoatsAreRockable:
@le717 Ah I'm sorry to hear that :(

le717:
That said, this is a major change I want to review. If you can hold off on merging for a bit that would awesome. :)

BoatsAreRockable:
Sure thing :)

le717:
I feel so bad I can't do anything right now. I really want to look through all this and get it merged, but without my laptop I can't test anything, and testing is half of review. :(

BoatsAreRockable:
@le717 Don't feel bad, it's not your fault at all! :)

BoatsAreRockable:
@le717 Thanks for your comment, merging this was not right thing to do. Of course no one likes to be criticised or told they're wrong, but when done in a polite and constructive way (as you did) it is greatly appreciated and I'm very happy to listen. If there any other issues like this now or in the future feel free to speak up; I won't be offended. :) 

I'll remove the merge commit from this branch and as you said, shortcuts can be reorientated once nwjs-013 is merged with master.

BTW I can't thank you enough for essentially mentoring me with this project. Your work here has been invaluable in making this a reality and helping me learn way more than I could of imagined in such a short space of time. I guess you've helped keep Boats Animator "a float"...

le717:
BTW, my laptop is in the mail on the way back. Let's hope the actually fixed it this time.

BoatsAreRockable:
@le717 that's great to here! :) I don't think I could last as long as you have without a computer.

le717:
Oh, I have had a computer.  I nabbed dad's Ultrabook and have been using it. I just have not had _my_ computer, thus all my development stuff. :)

le717:
\*sigh* Loosing my again for a third time because tech support creates problems. I am _so, so_ sorry to keep all this waiting. I really want to review/merge everything but I can't without my dev environment (and I can't really set it up on another computer). :(

BoatsAreRockable:
This should be rebased before it is merged.

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+        // Display the selected frame
+        var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
+        _displayFrame(imageID);
+
+      } else if (e.target.className === "frame-reel-img selected") {
+        addKeyframe("start", curSelectedFrame);
+
+      } else if (e.target.className.includes("keyframe")) {
+        // If the current start keyframe is double clicked it is removed
+        if (parseInt(e.target.id.slice(4)) === curStartKeyframe) {
```
le717: The radix parameter for `parseInt` needs to be given (hint: it's always `10`).

le717:
Initial thoughts are good, will have to look more deeply later (maybe even today).

le717:
Putting a total review of this PR on my Saturday schedule. I need to get this merged.

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
 
   // Preview a captured frame
   frameReelRow.addEventListener("click", function(e) {
-    if (e.target.className === "frame-reel-img") {
-      if (winMode !== "playback") {
-        switchMode("playback");
+    if (e.target.className.includes("frame-reel-img")) {
```
le717: `classList.includes` is not a valid method. You are looking for `classList .contains`. Fixing this may affect behavior (since the code is actually valid), so be sure to test everything afterward.

le717: See also: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+      // For an unselected, non-keyframe
+      if (!e.target.className.includes("selected")) {
```
le717: Yea, you might want to do a find-replace for all instances of `classList.includes`...

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+        e.target.classList.add("selected");
+        if (winMode !== "playback") {
+          switchMode("playback");
+        }
+
+        // Display the selected frame
+        var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
+        _displayFrame(imageID);
+
+      } else if (e.target.className === "frame-reel-img selected") {
```
le717: Try using [`Element.matches()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) here instead:

```js
} else if (e.target.matches(".frame-reel-img.selected")) {
```

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+    } else {
+      addKeyframe("start", curSelectedFrame);
+    }
+  });
 
-      // Display the selected frame
-      var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
-      _displayFrame(imageID);
+  // Remove the start keyframe
+  btnClearKeyframe.addEventListener("click", function(e) {
```
le717: Code structure tip: If you have a condition that `return`s to stop the rest of the function from running, you can eliminate the `else` clause.

```js
btnClearKeyframe.addEventListener("click", function(e) {
    if (!curStartKeyframe) {
      return false;
    }
    removeKeyframe("start");
  });
```

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+
+        // Highlight the clicked image
+        e.target.classList.add("selected");
+        if (winMode !== "playback") {
+          switchMode("playback");
+        }
+
+        // Display the selected frame
+        var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
+        _displayFrame(imageID);
```
le717: Go ahead and rewrite these into one line, as this is the only time `imageID` is used.

app/js/main.js
```diff
@@ -328,17 +333,55 @@ function startup() {
...
+        // If the current start keyframe is double clicked it is removed
+        if (parseInt(e.target.id.slice(4), 10) === curStartKeyframe) {
+          removeKeyframe("start");
+        }
       }
+    }
+  });
+
+  // Add a start keyframe
+  btnStartKeyframe.addEventListener("click", function(e) {
```
le717: IDK if you just missed the ESLint warning, but the `e` variable here and on line 376 should be removed.

app/js/main.js
```diff
@@ -657,14 +766,12 @@ function _videoPlay() {
     // There are no more frames to preview
     if (curPlayFrame >= totalFrames) {
       // We are not looping, stop the playback
-      if (!isLooping) {
-        videoStop();
-      } else {
+      videoStop();
```
le717: It looks like here you stop the playback before checking if we are looping, then resume the playback. Was this done for any particular reason?

app/js/main.js
le717: NVM my comment here. I'll remove the variables.

le717:
Review done, code looks good! Biggest concern is definitely the `classList.includes` usage, which could alter behavior

le717:
Wait a second, I just realized I have been confusing `classList` and `className`. My bad! What threw me off was `className.includes` is just [`String.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) (something new in ES6 I did not know about), while `classList` exposes it's own set of APIs for classes. IDK, I still think changing those lines to use  `classList.contains` would be a good idea.

le717:
I went ahead and made the changes I pointed out and did a force push to rebase part of the branch. You'll need to re-checkout the branch now. The question regarding stopping the playback before resuming looping still stands, and I I still need to merge this with `master` before it can be merged.

BoatsAreRockable:
Thanks @le717 for reviewing and fixing this!

It's been a while so I'm slightly rusty with my exact reasons for stopping the playback and resuming the looping. However it appears I might have possibly have done so in order to not have to repeat this if statement in both `_videoPlay()` and `videoStop()`:
~~~~
// Reset playback
  if (curStartKeyframe) {
    curPlayFrame = curStartKeyframe - 1;
  } else {
    curPlayFrame = 0;
  }
~~~~

Looking at It seems quite a trivial reason to have to recall previewCapturedFrames! I will change this.

BoatsAreRockable:
There are a couple more things I'd I need to fix on this branch before merging begins:

Firstly, I've found a bug where playback does not stop at the last frame. Replicate it by:
1. Set the last frame captured as a keyframe
2. Remove the keyframe
3. Press play

Another bug which seems to have appeared post review:
1. Set a frame as a keyframe by doubling clicking its thumbnail in the frame reel
2. Clicking it a 3rd time should remove the keyframe, however this no longer works.

BoatsAreRockable:
Fixed one bug but still have another to go. I'll try and look at it tonight.

BoatsAreRockable:
@le717 unfortunately I didn't get round to fixing the other bug and I'm away for the next week. Feel free to fix and then merge, otherwise this will have to wait.

le717:
I don't exactly know what the bug is. I mean, I followed your steps and get the same result, but I don't understand how it is a bug (I need to do the same on `master` and see what's up).

BoatsAreRockable:
Hi @le717,

Just to clarify, do you mean you do not understand why I think this is a bug or why this bug is occuring?

le717:
I don't understand the bug at all. I don't know what the proper behavior is vs the current, broken (according to you) behavior. That is all.

BoatsAreRockable:
@le717 I will come back to this tomorrow!

BoatsAreRockable:
I'm finding this confusing to explain, but hopefully this makes sense:

####Current broken behaviour:

1. Set the last frame captured as a keyframe
2. Remove the keyframe
3. Press play
4. Playback will not stop properly and console errors will be generated. The pause button will not be changed to a play button in the control panel and no frame has a white selection outline around it in the frame reel.

####Correct behaviour:
The correct behaviour for this should be that selecting the last frame captured and pressing play will result in playback technically starting but instantly stopping as there are no more frames to preview. The last frame captured should have a white outline around it.

Pressing the playback a second time in this instance will result in playback beginning from the first frame captured.

I acknowledge the correct behaviour doesn't really make total sense, but that's how it should be at the moment.

BoatsAreRockable:
I've finished updating this to work with the updated onion skin slider. The behaviour described in #71 has been fully implemented.

le717:
I'm playing with this right now. :)

le717:
OK, so I've gone though and changed it to single click. Behavior is much more intuitive and straight-forward now. Hopefully I've not broken anything in the process. However, for whatever reason, onion skin is still broken. I've traced down two things (there may be more). I'll list them and let you figure it out.

* When I add a keyframe, `#onion-skinning-frame` is not updated with the selected frame.
* It would appear that the opacity change is doing nothing. It looks like the preview window is covering up the onion skin preview completely. However, I can't find anything that changes it.
* The fact this started this morning, before my changes, and only worked for a few minutes, makes this even more confusing. :\

BoatsAreRockable:
That is very strange indeed, opacity change is working for me and never was broken. 

Without the blue outline I don't really understand how the user can tell the current location of the key frame.

Also the key frame system is now broken:
* When a user switches from playback to capture mode the keyframe is always set to the last frame captured - this means moving the onion skin slider to the left will always have the same effect as sliding it right which is the very behaviour this PR was supposed to change!
* When a user begins playback from a keyframe, the current keyframe increments with each frame displayed. The keyframe system is supposed to allow you the begin playback from a set location multiple times and therefore should only change when the user explicitly says so.
* When the last frame captured is reached in the frame reel, the play/pause and stop buttons do not begin playback from the beginning of the frame reel because a keyframe has been set to the last frame. 

I have to admit I am rather confused with the changes you have made to keyframe system. I am not entirely sure that we both have the same idea in our minds of how it is supposed to work!

le717:
I am terribly sorry. I was seeing it as more of an "advanced" form of frame selection... or something. IDK what, exactly. I've reset my changes and restored your code. I was trying to align it with Dragonframe's behavior, but I found it DF does not have this functionality so I was confused. The concept of "keyframes" is foreign to brickfilming (it's used in 3D animation), so that was throwing me and rioforce off.

That said, I've made two minor modifications after talking this over with rioforce:
* The "Add keyframe" button is confusing, so it was removed
* You should not be able to remove a keyframe by clicking it again

Here was the behavior rioforce said this should probably exhibit, again, since this is seemingly new behavior.

```
1.	Click on frame to make it gray
2.	Press play, it plays from frame to end and stops
3.	If looping, it goes to the beginning of the timeline
4.	If clicked another frame, it becomes gray
5.	Rinse and repeat

1.	Click on frame twice to make it blue
2.	Press play, it plays from frame to end and goes back to the frame
3.	If looping, it goes back to the blue frame then the end and the blue frame, etc etc.
4.	Press the “X” to remove the blue off the frame
5.	If clicked another frame once while a frame is blue, keep frame blue and still start from blue frame
```

We meet all this but `still start from blue frame` in the second half. Basically, when a "keyframe" is set, it overrides normal playback behavior, and the "gray" normal selection becomes a frame previewer.

Does this make sense or are both of us still confused?

le717:
OK, I found the onion skinning behavior. It doesn't act as I nor rioforce expected, which is why I thought it broke. I'll write up how I was expecting it to work soon. Let's just start with the above issue and get it figured out.

BoatsAreRockable:
Yes that makes sense and rioforce has got what I had in mind. I am of course open to a better and or radically different solution and since you two found this feature confusing, I would imagine the average user would find it even more so!

I'll hold off merging this until I see what you and rioforce were expecting of this. I haven't used Dragonframe enough to understand how this type of thing is handled there but I'll try and get round to downloading a trial.

BoatsAreRockable:
Also since I'd like to push forward with things I'm going to push this back to 0.8.x and krank out 0.7.4.

rioforce:
> Yes that makes sense and rioforce has got what I had in mind. I am of course open to a better and or radically different solution and since you two found this feature confusing, I would imagine the average user would find it even more so!

Are you speaking of the onion skinning or the playback-from-a-location feature? I think that the playing back from the location is a good idea, and returning to that set location is a good idea that not even Dragonframe has. Dragonframe allows the playback to start at the selected frame, but it will stop at the end of the timeline when the sequence is over and return to the beginning of the timeline if play is pressed again.

I took this quick video of the onion skinning functionality of Dragonframe earlier in case you need some help understanding how it's supposed to work. I'm afraid that I was super vague when I first described it all those ages ago. I'll need to check to see how the current functionality works, but for now, here's the video link if you need it. 😉 

https://www.youtube.com/watch?v=JiuaDUKF-4Q

le717:
The onion skinning behavior in question is as so:

**Current**
* When a frame is blue, you have to click live view to trigger onion skin.

**Expected**
* When a frame is blue, you can drag the slider and it will begin onion skinning

**Current**
* When a frame is blue, dragging the slider to the right overlays the last captured frame
* When a frame is blue, dragging the slider to the left overlays the blue frame

**Expected**
* When a frame is blue, dragging the slider to the right does nothing*
* When a frame is blue, dragging the slider to the left overlays the blue frame

\* This is how DF works, but it may be alright to keep the current behavior. We can have animators try it and tell us.

The first behavior is the most troublesome. The remaining code, skinning over the last captured when no frame is blue, works as expected.

BoatsAreRockable:
Thanks for the explanations.. :) I'm happy to rename the keyframe feature if anyone has suggestion for a clearer name for it.

In regards to onion skinning, while different from Dragonframe, would the following be suitable behaviour?

| Action | Capture mode | Playback mode |
| - | - | - |
| Slide left | If blue frame, onion skin with it, otherwise use last frame | If blue frame, onion skin with it, otherwise do nothing |
| Slide right | Onion skin with the last frame captured | Do nothing

Also when playback occurs should onion skinning be paused?

Sorry for my continued confusion and please do correct me if I still haven't quite got this!

le717:
You're fine. This is getting lots of discussion because we want it to right and we are trying to figure out what "right" it. :)

OK, let's backtrack. Let's leave the the slider alone ATM and focus on this one part right here:

> **Current**
> When a frame is blue, you have to click live view to trigger onion skin.
> **Expected**
> When a frame is blue, you can drag the slider and it will begin onion skinning

This is probably the biggest behavior issue right now. I do not want to have to mark a frame then go _all the way_ back to the live button, click it, _then_ begin onion skinning. I should be able to mark a frame and _immediately_ begin. I've recorded a quick video of the current behavior, which, BTW, is not expected or intuitive.

https://www.youtube.com/watch?v=I9PtzTU4M6s

Does what I am saying and expecting here make sense?

BoatsAreRockable:
@le717 I've just made an update to the onion skinning behaviour. It now operates in both playback and capture modes.

le717:
So you know how earlier this year I couldn't work on this because my laptop died?

Well, it died again. This time it's overheating and won't stay on. It's possible I'll be buying a new laptop.

Soooooo... yea. Sorry about that. 😞

BoatsAreRockable:
Oh no not again! I'm sorry to hear that 😞

PR #125, opened 2016-03-01
Improve statusbar

BoatsAreRockable:
I've made the status bar more compact and less "wordy":

Before:
beforestatusbar

After:
statusafter

I've also changed a couple of things that were bugging me:
#### Renamed curFrame
Firstly I've renamed the variable `curFrame` to `totalFrames`. This was getting really confusing for me because `curFrame` didn't really reflect the variable's purpose (which is to provide the total number of frames captured). Also the variables `curSelectedFrame` and `curPlayFrame` were similarly named to `curFrame` but act quite differently to it since they change value relatively frequently while `curFrame` is more of a constant. In addition, the variable `statusBarCurFrame` was, despite its name, often not equal to var `curFrame`

#### Changed frame number when in capture mode
When in capture mode the frame number in the status bar is now 0. It was previously equal to the number of frames captured (or the old `curFrame`) which was confusing because the last frame captured would also display a frame number of `curFrame`.

There is a debate to be had that instead of 0 the frame number in capture mode should one more than the frame total (ie 19 in the example above) - which is how Dragonframe works. However, I personally find this quite confusing and in this proposed form the status bar would say `frame 19 of 18` which does sound rather clumsy...

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -165,9 +165,9 @@
 
     <div id="statusBar">
         <ul>
-            <li>Frame <span id="currentFrame">0</span> of <span id="num-of-frames">0</span></li>
-            <li id="currentFrameRate"><span></span> FPS</li>
-            <li class="no-pipe" id="currentMode"><span></span> mode</li>
+            <li>Frame <span id="current-frame">0</span> of <span id="num-of-frames">0</span></li>
+            <li id="current-frame-rate"><span></span> FPS</li>
+            <li class="no-pipe" id="current-mode"><span></span> mode</li>
             <li class="no-pipe" id="statusBar-credits"><a href="#" onclick="gui.Shell.openExternal('https://github.com/BoatsAreRockable/animator')">&copy; Charlie Lee 2016</a></li>
         </ul>
     </div>
--- a/app/animator.html
+++ b/app/animator.html
@@ -165,8 +165,7 @@
 
     <div id="statusBar">
         <ul>
-            <li id="currentFrame">Current frame: <span>0</span></li>
-            <li id="num-of-frames"><span>0 frames</span> captured</li>
+            <li>Frame <span id="currentFrame">0</span> of <span id="num-of-frames">0</span></li>
             <li id="currentFrameRate"><span></span> FPS</li>
             <li class="no-pipe" id="currentMode"><span></span> mode</li>
             <li class="no-pipe" id="statusBar-credits"><a href="#" onclick="gui.Shell.openExternal('https://github.com/BoatsAreRockable/animator')">&copy; Charlie Lee 2016</a></li>
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -54,10 +54,10 @@ var width  = 640,
     audioToggle  = document.querySelector("#audio-toggle"),
 
     // Status bar
-    statusBarCurMode   = document.querySelector("#currentMode span"),
-    statusBarCurFrame  = document.querySelector("#currentFrame"),
+    statusBarCurMode   = document.querySelector("#current-mode span"),
+    statusBarCurFrame  = document.querySelector("#current-frame"),
     statusBarFrameNum  = document.querySelector("#num-of-frames"),
-    statusBarFrameRate = document.querySelector("#currentFrameRate span"),
+    statusBarFrameRate = document.querySelector("#current-frame-rate span"),
 
     // Export frames
     frameExportDirectory  = null,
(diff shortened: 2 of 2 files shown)


le717:
Visually looks nice, will look over code and review the best I can on mobile ASAP.

app/animator.html
```diff
@@ -165,8 +165,7 @@ <h2 id="confirm-title"><i class="fa fa-exclamation-triangle"></i> Confirm</h2>
 
     <div id="statusBar">
         <ul>
-            <li id="currentFrame">Current frame: <span>0</span></li>
-            <li id="num-of-frames"><span>0 frames</span> captured</li>
+            <li>Frame <span id="currentFrame">0</span> of <span id="num-of-frames">0</span></li>
```
le717: Go ahead and change this ID to `current-frame`.

app/js/main.js
```diff
@@ -383,7 +385,7 @@ function updateFrameReel(action, id) {
         // Update frame preview selection
         if (curSelectedFrame) {
             _removeFrameReelSelection();
-            document.querySelector(`.frame-reel-img#img-${id - 1}`).classList.add("selected");
+            _addFrameReelSelection(id - 1);
```
le717: Good catch!

app/js/main.js
```diff
@@ -336,6 +336,8 @@ function _removeFrameReelSelection() {
 function _addFrameReelSelection(id) {
     "use strict";
     document.querySelector(`.frame-reel-img#img-${id}`).classList.add("selected");
+    curSelectedFrame = id;
+    statusBarCurFrame.innerHTML = id;
```
le717: I'm not sold on these two lines. The method is called `_addFrameReelSelection`. Ideally, methods only interact with what they say they do. Now, this method says on the tin we only alter the frame reel, but it actually also edits the status bar.

Fun challenge for you: devise how this should be fixed on your own instead of me telling you. Let's see how much you are learning. :)

le717:
> Firstly I've renamed the variable curFrame to totalFrames. This was getting really confusing for me because curFrame didn't really reflect the variable's purpose (which is to provide the total number of frames captured).

+1

> Also the variables curSelectedFrame and curPlayFrame were similarly named to curFrame but act quite differently to it since they change value relatively frequently while curFrame is more of a constant. In addition, the variable statusBarCurFrame was, despite its name, often not equal to var curFrame

So do these need to be renamed too?

> There is a debate to be had that instead of 0 the frame number in capture mode should one more than the frame total (ie 19 in the example above) - which is how Dragonframe works. However, I personally find this quite confusing and in this proposed form the status bar would say frame 19 of 18 which does sound rather clumsy...

That does sound odd but 0 might sound worse. What does other software do in this matter? Also, don't we update the second value in capture mode? Then why not, while in capture mode, update both values to be exactly the same and always remain in sync (Frame 19 of 19, Frame 20 of 20, etc).

BoatsAreRockable:
> So do these need to be renamed too?

No these have names that make sense :)

> What does other software do in this matter?

Well:

#### Dragonframe in Live view
dragonframe

* The frame num at the bottom in live view is one more than the total number of frames captured.
* The number at the top is also one more than the num of frames captured. It remains constant during playback.

#### iStopMotion
##### In playback mode:


##### In live view:


A frame number is simply not shown in capture mode in iStopMotion.




> Also, don't we update the second value in capture mode? 

The 2nd number is updated when a picture is taken or deleted. When in capture mode and before a picture is taken saying `frame 19 of 19` doesn't make sense to me because if I've only taken 18 pictures why does the program say there's 19 in total. Perhaps this makes more sense to seasoned Dragonframe users, but I think it's weird!

@le717 I've decided after looking at the other programs that (with 18 pics captured and in capture mode) instead of saying `Frame 0 of 18` the status bar should say `frame 19 of 18`. Sure it's clumsy but it's chronological.

Also thanks for reviewing my code here, I'll push a commit here soon :)

le717:
Anytime you have this ready, I'm ready. :)

BoatsAreRockable:
@le717 I've just pushed some changes. The status bar frame number is updated using a new function called `_updateStatusBarCurFrame` (arguably a bit of a long name). In live mode the status bar now displays one more than the total number frames captured (ie frame 19 of 18).

le717:
Code looks good. We now suddenly have a merge conflict. Follow the directions GitHub provides (see image) and let me know when you've got that done.

BoatsAreRockable:
@le717 the conflicts have been restored!

le717:
No dude, you _resolve_ conflicts, not bring them back. :P

BoatsAreRockable:
@le717 ah you caught me before I edited it! ;)

PR #126, opened 2016-03-02
Add new utils module + various fixes

le717:
(Yes, this is all me. I have been using [Codeanywhere](https://codeanywhere.com/) this morning.)

I've noticed there are a few set of methods we reuse over and over. Thus, per DRY principles, we should move them into a common place. Thus that is what I've done. This module can and likely will be expanded.

Also included:
* Renamed `capturedFramesRaw` to just `capturedFrames`
* Fixed those menu registrations you mentioned
* Removed some `onclick` attributes (I discourage these as they break separation between markup/presentation and logic).
* Encapsulated the index JS in an [IFFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression). These help promote separation between presentation and logic.

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -166,6 +166,7 @@
         </ul>
     </div>
 
-    <script src="js/main.js" async></script>
+    <script src="js/utils.js"></script>  
+    <script src="js/main.js"></script>
 </body>
 </html>
--- a/app/animator.html
+++ b/app/animator.html
@@ -160,8 +160,8 @@
             <li class="no-pipe" id="currentMode"><span></span> mode</li>
             <li id="statusBar-dev" class="no-pipe">
                 <!--Debugging Options-->
-                <button onclick="dev()">Dev tools</button>
-                <button onclick="reload()">Refresh</button>
+                <button id="btn-open-dev-tools">Dev tools</button>
+                <button id="btn-reload-page">Refresh</button>
             </li>
         </ul>
     </div>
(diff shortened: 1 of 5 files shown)


le717:
Boats, your turn to review and merge when you think it is good. :)

le717:
Do note that we'll want to update the rest of the code base to use `openURL`. Without my laptop I am limited in my coding.

BoatsAreRockable:
@le717 Ah yes. Also I will review this later today - there's no obvious problems on first glance, but I'll give it a double check.

BoatsAreRockable:
@le717 Unfortunately this is giving me errors in the console :(
errors

This seems to be some sort of context issue. According to the NW.js wiki:

> You cannot use require('nw.gui') (to access the node-webkit's GUI API) from the Node's context, because there's no GUI outside of a window.

https://github.com/nwjs/nw.js/wiki/Differences-of-JavaScript-contexts#features-andlimitations-of-thenodes-context

I guess we can't use utils.js as a node module. How it is accessed will have to be rethought. A less pretty work around I can think of is to simple add `<script src="utils.js"></script>` to `animator.html`.

Otherwise your changes seem fine (although with the `gui` error I can't check them out). It's nice to see relics like `capturedFramesRaw` disappear!

le717:
Meh. Oh well, stuff like this happens. I've learned something new now. :smile:

I'll try to pull this up on Codeanywhere tomorrow and fix it. I know how I want to do it. You'll just have to make sure it works.

le717:
Changes pushed, should be work now.

BoatsAreRockable:
@le717 the only issue was that utils.js was not linked to on index.html. I've fixed that and will merge :)

Issue #91, opened 2015-12-06
Revised keyboard shortcuts


charlielee:
The repetitiveness of some aspects of stop motion animation make solid keyboard shortcuts a must for Boats Animator. A quirk of NW.js's menu API mean that with very little hassle the following shortcuts currently exist in BA:
- `Ctrl Z` = undo last frame
- `Ctrl C` = capture frame
- `Ctrl M` = return to launcher

These are not great for numerous reasons and need to revised to consider standard Mac and PC expectations amongst other things. Some suggestions I have for future and existing features are:

(note `NUM` refers to the keypad)
##### Projects
- [x] `Ctrl O` / `Cmd O` = open project
- [ ] `Ctrl S` / `Cmd S` = save project
- [ ] `Ctrl Shift S` / `Cmd Shift S` = save project as
- [x] `Ctrl N` / `Cmd N` = new scene
- [ ] `Ctrl Shift N` / `Cmd Shift N` = new take
- [x] `Ctrl W` / `Cmd W` = close window (and return to the launcher screen)
- [x] `Ctrl Q` / `Cmd Q` = quit program

_Project shortcuts will be added when #37 and #86 are implemented_
##### Controls
- [x] `Enter` / `NUM Enter` = Capture frame
- [x] `Ctrl Z` / `Cmd Z` / `Backspace` / `Delete` / `NUM *` = delete last frame captured
- [ ] `Ctrl Shift Z` / `Cmd Shift Z` / `Ctrl Delete` / `Cmd delete` = delete current selected frame in frame reel
- [x] `Ctrl M` / `Cmd M` / `M` = toggle audio
##### Playback
- [x] `Space` / `0` / `NUM 0` / `MediaPlayPause` = Play / pause
- [ ] `P` / `Num .` = Play from start
- [ ] `6` / `NUM 6` = Play last few captured frames
- [x] `8` / `NUM 8` = toggle looping
- [x] `Ctrl Right` / `MediaPrevTrack` = jump to the first frame on the framereel
- [x] `Ctrl Left` / `MediaNextTrack` = jump to the last frame on the framereel (ie stop)
  ##### Framereel
- [x] `left arrow` / `1` / `NUM 1` = back 1 frame on framereel
- [x] `right arrow` / `2` / `NUM 2` = forward 1 frame on framereel
- [x] `L` /  `3` / `NUM 3` = switch to live view
- [ ] `Ctrl +` / `Cmd +` = zoom into framereel
[...]

PR #130, opened 2016-03-05
Added mousetrap shortcuts to existing features

BoatsAreRockable:
Part of #91. I've also added the ability to begin playback from the 1st frame captured; this is particularly useful in playback mode. The other shortcuts I've implemented are checked off the list at #91.

In addition I've capitalised the menubar items' titles since I was already poking around with them.

I meant for this to be part of the 0.7.x series but I guess since this is ready it could go into 0.6.4.

### DIFF
--- a/app/animator.html
+++ b/app/animator.html
@@ -173,6 +173,7 @@
         </ul>
     </div>
 
-    <script src="js/main.js" async></script>
+    <script src="lib/mousetrap.js"></script>
+    <script src="js/main.js"></script>
 </body>
 </html>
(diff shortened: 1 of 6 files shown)


BoatsAreRockable:
@le717 This is ready for reviewing if possible :)

BoatsAreRockable:
@le717 I've reverted 54b2e52.

BoatsAreRockable:
NW.js 0.13 greatly improves its shortcut API. I think this should be redone using it.

le717:
If this PR is closed, it should be removed from the milestone. ;)

PR #137, opened 2016-03-31
Keyboard shortcuts using NW.js shortcut API

BoatsAreRockable:
This implements keyboard shortcuts for existing features #91.

I've redone the keyboard shortcuts using NW.js 0.13's shortcut API rather than using mousetrap.js #130.

This has been done in a way that should work well when custom shortcuts are implemented.

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -139,6 +139,7 @@ win.on("close", function() {
 
 function closeAnimator() {
     win.close(true);
+    nw.App.closeAllWindows();
 }
 
 function startup() {
@@ -893,21 +894,51 @@ function confirmSet(callback, args, msg) {
     "use strict";
     confirmText.innerHTML = msg;
     confirmContainer.classList.remove("hidden");
+    btnConfirmOK.focus();
+
+    removeShortcuts(mainKeys);
+    addShortcuts(confirmKeys);
+    editMenu.items[0].enabled = false;
+    captureMenu.items[0].enabled = false;
 
     function _ok() {
-        confirmContainer.classList.add("hidden");
         callback(args);
-        btnConfirmOK.removeEventListener("click", _ok);
+        _confirmSelect();
     }
 
     function _cancel() {
+        _confirmSelect();
+    }
+
+    function _confirmSelect() {
         confirmContainer.classList.add("hidden");
+
+        btnConfirmOK.removeEventListener("click", _ok);
         btnConfirmCancel.removeEventListener("click", _cancel);
+        
+        btnConfirmOK.removeEventListener("blur", _focusCancel);
+        btnConfirmCancel.removeEventListener("blur", _focusOK);
+
+        removeShortcuts(confirmKeys);
+        addShortcuts(mainKeys);
+        editMenu.items[0].enabled = true;
+        captureMenu.items[0].enabled = true;
     }
 
     // Respond to button clicks
     btnConfirmOK.addEventListener("click", _ok);
     btnConfirmCancel.addEventListener("click", _cancel);
+    
+    function _focusOK() {
+        btnConfirmOK.focus();
+    }
+
+    function _focusCancel() {
+        btnConfirmCancel.focus();
+    }
+
+    btnConfirmOK.addEventListener("blur", _focusCancel);
+    btnConfirmCancel.addEventListener("blur", _focusOK);
 }
 
 /**
@@ -1052,6 +1083,12 @@ var shortcuts = {
         if (totalFrames > 0) {
             btnLiveView.click();
         }
+    },
+    confirmOK: function() {
+        btnConfirmOK.click();
+    },
+    confirmCancel: function() {
+        btnConfirmCancel.click();
     }
 };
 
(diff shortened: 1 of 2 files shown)


app/js/main.js
```diff
@@ -916,20 +955,28 @@ function loadMenu() {
         helpMenu    = new nw.Menu(),
         debugMenu   = new nw.Menu();
 
+function loadMenu() {
+    "use strict";
+    if (process.platform === "darwin") {
```
le717: Rewrite to use a ternary:

```js
var controlKey = process.platform === "darwin" ? "command" : "ctrl";
```

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+    {key: "NumpadDivide", active: shortcuts.confirmCancel}
+];
+
+/**
+ * Choose an array of shortcuts to activate.
+ *
+ * @param {variable} shortcutArray Which array of shortcuts should be enabled.
+ *                                 eg mainKeys or confirmKeys.
+ */
+function addShortcuts(shortcutArray) {
```
le717: Can we move this add/remove mechanism to a separate script (`shortcuts.js`) and `require()` it into here (`file.js` would be a good model to follow)?

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+            window[shortcutArray[i].key + "Shortcut"] = new nw.Shortcut(option);
+            nw.App.registerGlobalHotKey(window[shortcutArray[i].key + "Shortcut"]);
+        }
+
+        enableShortcuts = true;
+        console.log("added shortcuts");
+    }
 }
+
+function removeShortcuts(shortcutArray) {
```
le717: Function is missing JSDoc.

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+var confirmKeys = [
+    {key: "Enter", active: shortcuts.confirmOK},
+    {key: "Escape", active: shortcuts.confirmCancel},
+    {key: "NumpadDivide", active: shortcuts.confirmCancel}
+];
+
+/**
+ * Choose an array of shortcuts to activate.
+ *
+ * @param {variable} shortcutArray Which array of shortcuts should be enabled.
```
le717: The text in the braces defines the type (string, array, object, etc) the variable is expected to it. In this case, it should be `{Array}`.

app/js/main.js
```diff
@@ -93,7 +93,11 @@ var width  = 640,
     mkdirp = require("./lib/mkdirp"),
 
     // Sidebar
-    btnDirectoryChange = document.querySelector("#sidebar #btn-dir-change");
+    btnDirectoryChange = document.querySelector("#sidebar #btn-dir-change"),
+
+    // Shortcuts
+    controlKey = null,
```
le717: Remove `controlKey` from the global scope. It is only used in `loadMenu()` and should remain in that scope only.

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+            // Options to apply to nw.Shortcut
+            var option = {
+                key : shortcutArray[i].key,
+                active : shortcutArray[i].active,
+                failed : function(err) {
+                    console.error(err);
+                }
+            };
+            window[shortcutArray[i].key + "Shortcut"] = new nw.Shortcut(option);
+            nw.App.registerGlobalHotKey(window[shortcutArray[i].key + "Shortcut"]);
```
le717: Should use string interpolation here and on line 1164.

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+
+/**
+ * Choose an array of shortcuts to activate.
+ *
+ * @param {variable} shortcutArray Which array of shortcuts should be enabled.
+ *                                 eg mainKeys or confirmKeys.
+ */
+function addShortcuts(shortcutArray) {
+    if (!enableShortcuts) {
+        for (i = 0; i < Object.keys(shortcutArray).length; i++) {
```
le717: Allow me to teach you about the `forEach` loop. :)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

This loop now becomes:
```js
shortcutArray.forEach(function(shortcut) {
var option = {
  key : shortcut.key,
  active : shortcut.active,
  failed : function(err) {
    console.error(err);
  }
});
// etc...
```

As for the one in `removeShortcuts`:
```js
shortcutArray.forEach(function(shortcut) {
  nw.App.unregisterGlobalHotKey(window[`${shortcut.key} Shortcut`]);
});
```

:)

app/js/main.js
```diff
@@ -1020,4 +1067,115 @@ function loadMenu() {
...
+            // Options to apply to nw.Shortcut
+            var option = {
+                key : shortcutArray[i].key,
+                active : shortcutArray[i].active,
+                failed : function(err) {
+                    console.error(err);
+                }
+            };
+            window[shortcutArray[i].key + "Shortcut"] = new nw.Shortcut(option);
```
le717: Why are we polluting the global `window` namespace? I am not seeing the reasoning here.

le717:
This PR is an excellent candidate for you to reformat the areas you edited to comply with #140. :)

PR #139, opened 2016-03-31
Removed dev tools and reload buttons

BoatsAreRockable:
This removes the dev-tools and refresh buttons that were in various locations. Nw.js 0.13's SDK builds provide access to dev tools via F12 and a "reload app" option accessed via right click.


le717:
I'll take a look at this in a few minutes. Just got home from work and am about to get my shower.

PR #144, opened 2016-04-02
Implement keyboard shortcuts using NW.js shortcut API

BoatsAreRockable:
@le717 Thank you for reviewing #137! I've pushed the changes you suggested and recreated this PR so that I can merge this with the nwjs-013 branch rather than nwjs-013-merge-master


BoatsAreRockable:
Will make some additional changes and then repoint this to 'master'

Issue #142, opened 2016-04-02
Newsfeed does not work and cannot leave launcher when app is exported using Web2Exe


BoatsAreRockable:
The newsfeed does not work when app is exported using Web2Exe. As a result "New Project" has no effect and the rest of the app can't be loaded. This issue does seem to be resolved by converting `newsfeed.js` to use `<script>` tags in index.html. The newsfeed works fine in the NW executable before exporting the app so I guess it's a Web2Exe issue.

The console error I'm getting is:
<img width="482" alt="capture" src="https://cloud.githubusercontent.com/assets/8095888/14226252/c6036468-f8d4-11e5-98d3-246d1e99018e.PNG">
(Which I can see because some fine chap suggested Web2Exe add support for SDK builds :P)

BoatsAreRockable:
This seems to be to do with one of the settings Web2Exe moves around in package.json after exporting. I shall work on isolating which one.

I also had a look at [nw-builder](https://github.com/nwjs/nw-builder), it has potential but does not currently work with NW.js 0.13+

BoatsAreRockable:
I've managed to get nw-builder to work with 0.13 using [this fork](https://github.com/evshiron/nw-builder).

le717:
Did you figure out what key is casing the issue with Web2Exe, and in your opinion, would it be better to switch to nw-builder? Why or why not?

BoatsAreRockable:
@le717 I didn't figure it out.

However I've just pushed #149 which changes the build method to use nw-builder. I would like to switch to nw-builder because it seems simpler than Web2Exe and eventually could be included as a node dependency in package.json.

le717:
Now that #150 has been merged, I presume this has been made null and can be closed?

BoatsAreRockable:
@le717 that is correct!

PR #145, opened 2016-04-02
Merge Nwjs-013 branch with master

BoatsAreRockable:
I believe that nwjs-013 is now stable enough to be merged to master. I believe everything works now, the only issue is #142 but I think that is a Web2Exe issue.


le717:
Behold! This PR can be merged again! I'll let you do the honors and update us to nwjs 0.13, and the 0.7.x series can officially kick off!

PR #147, opened 2016-04-05
Feature documentation

### LINKED ISSUE #146: Help documentation
It would be good for there to be help topics explaining the app's features as well as other useful information such as which cameras are supported and system requirements. Also if Boats Animator were to ever gain any kind of traction it make my life much easier to have help docs to direct people to!

I've been doing [some experiments](http://boatsanimator.readthedocs.org/en/064-docs) with Read the Docs and Sphinx (initially out of general interest rather than anything to do with this project). It seems relatively straight forward to setup and with a small bit of tweaking can accept Markdown files and CSS modifications. I think it could potentially be a solution to use.

Sphinx also handles versioning well and would encourage adding documentation at the same time as new features are added.

Other possible help solutions:
* Simply use this GitHub repo's Wiki
* Use a more traditional help authoring tool like [HelpNDoc](http://www.helpndoc.com/)

BoatsAreRockable:
This fixes #146. This adds detailed documentation about the program using Mkdocs. At the moment it can be viewed at: <http://boatsanimator.readthedocs.org/en/064-docs/>

The documentation I've written so far is relevant to 0.6.4. I would suggest when this PR is merged v0.6.5 is released in order to sync this up nicely with Read the Docs. Going forward new features should be documented as they are implemented.


le717:
Is it possible to move the shortcuts to the header, e.g, "Capture (Ctrl+C)"? I think it would be clearer that way. If you want, you can post a screenshot in a comment so I can see how it looks.

le717:
Also, to the technical, there are very few shortcuts in 0.6.x. I did not notice: is the current documentation tailored for 0.6 or 0.7?

BTW, after I merge this, there will be a change I want to make to the app in relation to these docs.

BoatsAreRockable:
@le717 my initial post: :wink: 
> The documentation I've written so far is relevant to 0.6.4. I would suggest when this PR is merged v0.6.5 is released in order to sync this up nicely with Read the Docs. Going forward new features should be documented as they are implemented.

To start with I think it would be good to at least have documentation for what has already been released.



> Is it possible to move the shortcuts to the header, e.g, "Capture (Ctrl+C)"?

The issue with doing this is that it breaks links to sections of the page (ie `main-window#capture-button` would become `main-window#capture-button-ctrl-c`).

As an improvement I've made the shortcuts admonitions:
docs admonition
 
For 0.7 I'd like to create a separate page listing all of the shortcuts.

BoatsAreRockable:
Also before this is implemented I'm debating whether the URL should be `animator.readthedocs.org` or `boatsanimator.readthedocs.org`..

le717:
boatsanimator, as the name of the program is "Boats Animator". ;)

BoatsAreRockable:
@le717 change it shall! 

On a kind of related note, I've been considering changing the GitHub repository's name too. Although I'm not sure if doing so would cause any issues (GitHub [claims not](https://help.github.com/articles/renaming-a-repository/)). If I were to change that it would be to either: `boats-animator`, `BoatsAnimator` or `boatsanimator`. (probably the first one)

le717:
I was going to bring up renaming it. I suggest Boats-Animator. Keep it pretty and in proper casing.

I'll comment on the revisions to the docs after lunch.

le717:
Shouldn't the docs be versioned 0.65 since that is the version they will first appear in?

> The issue with doing this is that it breaks links to sections of the page (ie main-window#capture-button would become main-window#capture-button-ctrl-c).

> As an improvement I've made the shortcuts admonitions:

Ah yes, that would wreck the link anchors. However, I'm not sold on the display. it just seems odd to have a large box with nothing but tiny, mono-spaced text with special styling in it. How would this look?

```md
## Undo Last Frame ##
**Shortcut: `<Ctrl+z>`, `<Cmd+z>`**
```

BoatsAreRockable:
@le717 I've updated the shortcuts to what you suggested. I think this looks better.

> Shouldn't the docs be versioned 0.6.5 since that is the version they will first appear in?
Good point, this branch is inaccurately named then I guess. Unlike Sphinx, Mkdocs doesn't appear to have any version number configuration setting. Read the Docs automatically creates a new version of the docs when one is released on GitHUb

le717:
If you think this is ready, go ahead and merge it.  Seeing as I am without my laptop for a third time, I am unable to do anything. That means you'll need to be decisive on the current changes and I'll take a passive view. :)

PR #149, opened 2016-04-07
Change build method to nw-builder

BoatsAreRockable:
This changes the build method from using Web2Exe to nw-builder. While the main nw-builder repo does not currently work with NW.js 0.13, [this fork](https://github.com/evshiron/nw-builder) does (or [my own](https://github.com/BoatsAreRockable/nw-builder)).

In addition this PR fixes #134.

Build instructions:
~~~
npm install evshiron/nw-builder -g
node build.js
~~~

**Note #145 should definitely be merged before considering this!**

Issue #134, opened 2016-03-15
Stop node-modules from being packaged


BoatsAreRockable:
For some reason it hasn't occurred to me before that the node-modules directory is incorrectly packaged when Web2Exe is run. Either some method to exclude the folder from Web2Exe should be used or I should result to temporarily removing node-modules when packaging the app.

le717:
Oh yup, that's how you blow up the release archives, for sure. :P

BoatsAreRockable:
@le717 Any objections to solving this by moving node_modules to the root folder `animator` rather `animator/app`?

PR #150, opened 2016-04-08
Switched build method to nw-builder

BoatsAreRockable:
Redone #149 in what I believe to be the correct manner. This changes the build method from using Web2Exe to nw-builder. Unlike #149 this is initially for v0.6.x and nw.js 0.12.3. 

Build instructions:
~~~
npm install evshiron/nw-builder -g
node build.js
~~~


le717:
What all does this change bring, other than 0.13 support? Smaller binaries,  better binary stability or debugging? Easier building? Will we have to switch back to the other way in the future for some reason?  Is this a temporary measure? Explain why you think we need this change.

(I don't have any real objections here, but answer the questions as if I do. These are questions you will face in one form or another. ;))

le717:
Also,  since this is for 0.6.x, I shall add it to the milestone.

BoatsAreRockable:
@le717 I've pushed some changes. 

> Explain why you think we need this change.

Reasons I think we should switch to nw-builder (other then 0.13 support):
* Easier build process (one command)
* Makes it easy to preview the app (using the `-r` argument)
* I don't like how Web2Exe pollutes the `package.json` file with extra stuff.
* When #134 is fixed this could be added as a node dependency.

le717:
Finish up 0.6.x, my man!

PR #152, opened 2016-04-23
Switched JavaScript linter to ESLint

BoatsAreRockable:
As it has ES6 support. In Brackets I'm using https://github.com/zaggino/brackets-eslint


le717:
I like it! You should really never use JSLint. ;) Plus, this gives me a reason to config VS Code to use ESLint like I've needed to do.

It may be better to use a [configuration file](http://eslint.org/docs/user-guide/configuring#configuration-file-formats) than editing each file. ;)

BoatsAreRockable:
@le717 I've created a config file. On the topic of linting, what are `/* global Buffer, process */` in `about.js` and `/* global Buffer, process, utils */` in `main.js` for and are they necessary? No error seem to show up when they are removed.

le717:
I just added two linting rules. They will aid in answering your question. Pull my commit then try removing the lines, looking up the rule's docs if needed ;)

BoatsAreRockable:
@le717 ah I see now! I've removed the lines and fixed some errors that showed up.

app/animator.html
```diff
@@ -9,7 +9,7 @@
     <link rel="stylesheet" href="css/animator.css">
 </head>
 
-<body onload="startup()">
```
le717: I am so happy to see this go. :smile:

app/js/utils.js
```diff
@@ -1,20 +1,18 @@
 const utils = (function() {
-    "use strict";
```
le717: I went ahead and reformatted this script to use 2 space.

le717:
LGTM (Looks good to me), merging.

PR #153, opened 2016-04-24
Add file copy and rename methods

### LINKED ISSUE #127: Add file rename method to file.js
Filing mainly as a reminder for me because we will need this method. Use [this code](https://github.com/le717/CodeTriangle.me/blob/fcdbbf85e2726254a842d3d16d8cd4255d4fb720/scripts/copy-js.js) as a reference.

le717:
Fixes #127. Just wanted to go ahead and add it in; one less thing to do later. I did not realize `fs.rename` was a thing until I had the copy code in place, so I left it in as a copy method. Do note the TODO in the copy method. You can go ahead and merge this if you'd like, or I can later when I eventually find the time to fill in that TODO (which will be after I finish setting back up my laptop and getting BA and everything else set back up).


BoatsAreRockable:
@le717 since there isn't anything urgently waiting for this to be implemented, I'm going to leave it for now. Best of luck with your continued laptop troubles!

le717:
This is ready to be merged. You can do or I can, doesn't matter.

PR #155, opened 2016-04-27
Implement keyboard shortcuts

BoatsAreRockable:
This implements keyboard shortcuts using the NW.js shortcut API. The shortcuts this adds are those which are checked off in #91.

After a great deal of re-factoring, all shortcut related code has been moved to `shortcuts.js` and the list of keyboard shortcuts that are used are stored in a separate json file: `default-shortcuts.json`. I have implemented this in a way that should accommodate custom keyboard shortcuts well in future.

Issue #136, opened 2016-03-31
Revise playback method


BoatsAreRockable:
@le717 in #112:
> I just figured out how to best playback the captured images.

>We currently display the video feed by drawing it to a `<canvas>`. Upon switching modes, we clear the canvas, pull the image data from `capturedFrames`, and draw it. Upon clicking play, we use `requestAnimationFrame`, throttled by the frame rate value, to loop through and draw the images to the canvas. When we are done and switch back to capture, clear canvas and restore the video stream.

>Downside: rAF is capped at 60 FPS, per browser limit and spec design choice, but that is likely acceptable.

The difference in playback performance in nwjs-0.13 I've experienced is a serious issue. As a result I'd like to make this a priority.

PR #156, opened 2016-04-30
Implement previous, next, first and last frame buttons

### LINKED ISSUE #141: Additional playback controls
As well as eventual keyboard shortcuts there should be buttons in the interface for previewing the previous and next frame on the framereel as well as buttons for jumping to the first and last frame captured. (basically buttons for `|<`, `<`, `>` and `>|`)

A play from start function should exist. This wouldn't have a button in the interface but could be activated by shortcut `p` or `numpad .`

BoatsAreRockable:
This PR fixes #141 by implementing previous, next, first and last frame buttons. In addition it moves the audio toggle from the sidebar to the playback controls area:

issue-141

From the picture above you can see that this does make the playback controls area a bit cluttered. As a result, when #155 is implemented, I think it would be good if the "next" and "previous" buttons were removed and their functionality be made available purely via keyboard shortcuts.

Playback controls area without previous and next buttons:
issue-141-2

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -53,8 +53,10 @@ var width  = 640,
     inputChangeFR = document.querySelector("#input-fr-change"),
 
     // Audio
-    captureAudio = new Audio("audio/camera.wav"),
-    audioToggle  = document.querySelector("#audio-toggle"),
+    captureAudio       = new Audio("audio/camera.wav"),
+    btnAudioToggle     = document.querySelector("#btn-audio-toggle"),
+    btnAudioToggleIcon = document.querySelector("#btn-audio-toggle i"),
+
 
     // Status bar
     statusBarCurMode   = document.querySelector("#current-mode span"),
@@ -255,18 +257,20 @@ function startup() {
 
   // Preview one frame to the right on framereel
   btnNext.addEventListener("click", function() {
-    if (curSelectedFrame !== totalFrames && winMode === "playback") {
-      videoJumpTo(curSelectedFrame + 1);
-    } else {
-      btnLiveView.click();
+    if (curSelectedFrame) {
+      if (curSelectedFrame !== totalFrames) {
+        videoJumpTo(curSelectedFrame + 1);
+      } else {
+        btnLiveView.click();
+      }
     }
   });
 
   // Preview one frame to the left on framereel
   btnPrevious.addEventListener("click", function() {
-    if (curSelectedFrame > 1 && winMode === "playback") {
-      videoJumpTo(curSelectedFrame - 1);
-    } else if (winMode === "capture") {
+    if (curSelectedFrame > 1) {
+        videoJumpTo(curSelectedFrame - 1);
+    } else  if (winMode === "capture") {
       switchMode("playback");
       videoJumpTo(totalFrames);
     }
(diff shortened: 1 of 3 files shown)


le717:
Does this need to be merged before #136 is implemented?

BoatsAreRockable:
@le717 Ah that was not something I had considered. I suppose so.

le717:
In terms of behavior and UI, it works and it seems you have handled all the cases. I'll look at the code part right after dinner.

TBH, I would not have moved the audio toggle there. I've been meaning to file this issue, but it should just be moved into the Capture menu. It does not need to be present in the UI.

BoatsAreRockable:
> I would not have moved the audio toggle there.
> It does not need to be present in the UI.

@le717 Good point. I'll move it.

app/js/main.js
```diff
@@ -543,7 +528,7 @@ function _toggleOnionSkin(ev) {
  */
 function audio(name) {
     "use strict";
-    if (btnAudioToggle.classList.contains("active")) {
+    if (playAudio === true) {
```
le717: Should be a truthy check:

`if (playAudio) {`

app/js/main.js
```diff
@@ -696,7 +681,6 @@ function previewCapturedFrames() {
     // Begin playing the frames
     isPlaying = true;
     playBackLoop = setInterval(_videoPlay, (1000 / frameRate));
-    console.info("Playback started");
```
le717: Why was this message removed?

app/js/main.js
```diff
@@ -46,11 +46,15 @@ var width  = 640,
     btnLoop       = document.querySelector("#btn-loop"),
     playback      = document.querySelector("#playback"),
     btnPlayPause  = document.querySelector("#btn-play-pause"),
+    btnNext       = document.querySelector("#btn-next"),
```
le717: Change ID selectors to following format for clarity:

`#btn-frame-[location]` (e.g. `#btn-frame-next`)

app/js/main.js
```diff
@@ -244,13 +248,51 @@ function startup() {
...
+        btnLiveView.click();
+      }
+    }
+  });
+
+  // Preview one frame to the left on framereel
+  btnPrevious.addEventListener("click", function() {
+    if (curSelectedFrame > 1) {
+        videoJumpTo(curSelectedFrame - 1);
+    } else  if (winMode === "capture") {
```
le717: Nit: remove extra space between else and if.

app/js/main.js
le717: Do the same the variable names too (e.g., `btnFrameNext`)

app/js/main.js
```diff
@@ -484,7 +528,7 @@ function _toggleOnionSkin(ev) {
  */
 function audio(name) {
     "use strict";
-    if (audioToggle.checked) {
+    if (playAudio === true) {
```
le717: Same truthy check here.

app/js/main.js
```diff
@@ -563,15 +607,31 @@ function videoPause() {
...
+  curPlayFrame = 0;
+  console.info("Playback stopped");
+}
+
+/**
+ * Pause playback and view a specific frame in the preview area.
+ *
+ * @param {Integer} id The frame ID to preview.
+ */
+function videoJumpTo(id) {
```
le717: Try `function _displayFrame(id) {` for clarity.

le717:
I've gone though all I can ATM, will review more later. Fix up those changes for now. ;)

app/js/main.js
BoatsAreRockable: @le717 Ah this is was an accident!

app/js/main.js
```diff
@@ -964,6 +1023,20 @@ function loadMenu() {
       modifiers: "ctrl",
     }));
 
+  captureMenu.append(new nw.MenuItem({ type: "separator" }));
+
+  captureMenu.append(new nw.MenuItem({
+    label: "Play capture sounds",
+    click: function() {
+      playAudio = !playAudio;
+      notifyInfo(`Capture sounds ${playAudio ? "enabled" : "disabled"}`);
```
le717: I don't think we need this line. The checkbox in the Capture menu is enough indication.

le717:
Alright, review done. I like it. When you finish up the review changes, let me know and I'll merge it.

BTW, I plan on fixing the playback as the next change, then keyboard shortcuts, then we'll call it v0.7.1 and you can promote it on BiM. We can't mere

app/js/main.js
BoatsAreRockable: @le717 The only thing about removing this is that it means there is visual response when this called via `ctrl+m`.

BoatsAreRockable:
@le717 Thank you for reviewing this, I've just finished making the review changes.

The v0.7.1 plan sounds good. You may have noticed that I haven't really promoted 0.7.0 since it didn't bring any major visual changes (and because of the playback issue) but 0.7.1 will get the full blog and BiM treatment!

app/js/main.js
le717: In that case, its fine. Multiple notification support will help things.

PR #157, opened 2016-04-30
Fix #151 and make minor feed UI adjustments

### LINKED ISSUE #151: Links in newsfeed excerpts do not work
HTML anchor tags should show up in the launcher's news feed.

More details:
> [Links] do not currently show in the content of posts. `post.excerpt` needs to be replaced with `post.content` and then a method to convert these anchors to use `utils.url` rather than loading the website within the app needs to be added.

le717:
Remember, dates are formatted per locale format, so don't freak over the American format. ;)

feed

### DIFF
--- a/app/js/newsfeed.js
+++ b/app/js/newsfeed.js
@@ -12,7 +12,7 @@ module.exports = {};
      */
     function openNews(e) {
         if (e.target.classList.contains("post-title")) {
-            nw.Shell.openExternal(e.target.parentElement.dataset.url);
+            utils.openURL(e.target.parentElement.dataset.url);
         }
     }
 
@@ -68,7 +68,7 @@ module.exports = {};
     function _compilePostHTML(post) {
         return `<div class="post" id="post-${post.id}">
     <header>
-        <a class="post-link" href="#" data-url="${post.url}"><h2 class="post-title">${post.title}</h2></a>
+        <h2 class="post-title"><a class="post-link" href="#" data-url="${post.url}">${post.title}</a>
         <p class="post-date">${post.date}</p>
     </header>
     <div class="post-content"><p>${post.excerpt}</p></div>
--- a/app/js/newsfeed.js
+++ b/app/js/newsfeed.js
@@ -11,8 +11,8 @@ module.exports = {};
    * @param {Object} e - An MouseEvent object.
    */
   function openNews(e) {
-    if (e.target.classList.contains("post-title")) {
-      utils.openURL(e.target.parentElement.dataset.url);
+    if (e.target.classList.contains("post-link")) {
+      utils.openURL(e.target.dataset.url);
     }
   }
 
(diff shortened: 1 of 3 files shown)


BoatsAreRockable:
> Dates are formatted per locale format

Hmm you say that, but interestingly for me the American date format still shows up! This was actually an issue for me before this PR too (which I kept forgetting to report).

Also #151 does not seem to be fixed. I think I was a bit vague in the issue description but essentially I meant that anchors do not currently show in the content of posts. `post.excerpt` needs to be replaced with `post.content` and then a method to convert these anchors to use `utils.url` rather than loading the website within the app needs to be added.

le717:
... The dates are still messed up? That's _weird_. I'll look into it.

Yea, that would have been nice detail to have in the issue. :P I thought you meant the header links. TBH, I don't know if we should do that. It's an excerpt for a reason. It's supposed to lead you to the full post. If you wrote a really long post, complete with links, bold/italic text, and images, then all that would be pulled into the feed and I'd have to filter it all out as well as truncate it somewhere. I've done it before, but it was a pain to do. I personally would rather keep using excerpt.

BoatsAreRockable:
@le717 I've just fixed a bug that prevented the post title links from working. Also I'm happy to leave the excerpt if you think it'll be painful to do.

app/js/newsfeed.js
```diff
@@ -68,7 +68,7 @@ module.exports = {};
   function _compilePostHTML(post) {
     return `<div class="post" id="post-${post.id}">
     <header>
-        <h2 class="post-title"><a class="post-link" href="#" data-url="${post.url}">${post.title}</a>
+        <h2 class="post-title"><a class="post-link" href="#" data-url="${post.url}">${post.title}</a></h2>
```
le717: This is what happens when I code tired. Thanks!

le717:
[Here is my code](https://github.com/LEGOUniverseNews/LUN-Patcher/blob/gh-pages/js/newsfeed.js#L33-L71) that does the filtering I told you about (visible [here](http://legouniversenews.github.io/LUN-Patcher/)). It works but I don't really like it (as I explained with my pov of an excerpt's purpose).

BoatsAreRockable:
@le717 thanks for sharing the filter code. I think we should leave things how they are and not implement inline post links.

BoatsAreRockable:
@le717 I suppose fixing the post dates is out of the scope of this PR. Also I'm going to close #151 unless you want to object.

One thing I've noticed with this PR is that the "Recent News" title is noticeably larger than "Projects". Is this intentional?

Also the bug fixes I previously made here changed the appearance of the newsfeed dates from the screenshot above, so you might want to check it out again if you haven't already.

le717:
>  I suppose fixing the post dates is out of the scope of this PR. 

Sadly, yes, there is no way to get the computer's locale (only nwjs's locale), which is needed to properly order the numbers. We can always switch the naming the months, which will help.

> One thing I've noticed with this PR is that the "Recent News" title is noticeably larger than "Projects". Is this intentional?

Unintentional, now fixed.

> Also the bug fixes I previously made here changed the appearance of the newsfeed dates from the screenshot above, so you might want to check it out again if you haven't already.

Still looks good. +1

Merge when ready.

BoatsAreRockable:
I didn't mean to squash merge this but I guess it doesn't really make a difference to things.

PR #159, opened 2016-05-01
Add frame reel image numbers

BoatsAreRockable:
This adds frame numbers to corner of each frame in the framereel:

frame-nos

I must admit I'm torn whether this is actually beneficial though. If this is implemented I would suggest it is made an optional setting.


le717:
No, this should be implemented, but it should be done in a different visual manner (think closer to Dragonframe's filmstrip).

BoatsAreRockable:
@le717 I've adjusted the frame numbers so that they now located above each frame:

frame reels nos

app/css/animator.css
```diff
@@ -302,13 +302,17 @@ a {
   top: -3px;
   overflow-x: scroll;
   background-color: #171717;
+  height: 6.2em;
 }
 
 #area-frame-reel table {
-  border-spacing: none !important;
+  border-spacing: 0 !important;
```
le717: If only we did not need this `!important`...

PR #168, opened 2016-05-15
Revised playback method

### LINKED ISSUE #136: Revise playback method

### LINKED ISSUE #158: Playback does not occur at the right speed
This is quite a major bug! And has been present since the update to nw.js 0.13. Using a metronome I've found the playback speed to be approximately 4 times slower than it should be. 

In #136 using `requestAnimationFrame` is suggested as a replacement for playback in the preview area. I can confirm removing the line `playback.setAttribute("src", capturedFrames[curPlayFrame]);` does greatly improve the performance of things..

I suppose the culprit is `setInterval` not performing as well as it should. Hopefully this can be optimised or an alternative method can be used.

le717:
Fixes #136 and fixes #158. This restores appropriate playback speeds and improves playback overall using the method I explained in #136. It also performs some optimization work previously only done in my WIP, unpublished (and now lost) camera resolutions/selection code.

Additionally contains the following fixes:
* Make `videoStop()` actually stop playback by restoring a call to `videoPause()`, broke in 75f9dafbf78164d18d40a1c20808b5d8640ef334
* Correct strange `NaN` check in FPS changing (the condition will never validate, `console.log(NaN ? true : false); // false`)

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -325,26 +325,18 @@ function startup() {
     }
   });
 
-    // Preview a captured frame
-    frameReelRow.addEventListener("click", function(e) {
-        if (e.target.className === "frame-reel-img") {
-            // Remove previous selection
-            _removeFrameReelSelection();
+  // Preview a captured frame
+  frameReelRow.addEventListener("click", function(e) {
+    if (e.target.className === "frame-reel-img") {
+      if (winMode !== "playback") {
+        switchMode("playback");
+      }
 
-            // Highlight the clicked image
-            e.target.classList.add("selected");
-            if (winMode !== "playback") {
-                switchMode("playback");
-            }
-
-            // Display the image and update all the necessary values
-            var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
-            playback.setAttribute("src", capturedFrames[imageID - 1]);
-            curPlayFrame = imageID - 1;
-            curSelectedFrame = imageID;
-            _updateStatusBarCurFrame(imageID);
-        }
-    });
+      // Display the selected frame
+      var imageID = parseInt(e.target.id.match(/^img-(\d+)$/)[1], 10);
+      _displayFrame(imageID);
+    }
+  });
 }
 window.onload = startup;
 
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -11,8 +11,8 @@ var width  = 640,
 
     // The various HTML elements we need to configure or control.
     preview     = document.querySelector("#preview"),
-    video       = document.createElement("video"),
-    canvas      = document.createElement("canvas"),
+    playback    = document.querySelector("#playback"),
+    context     = playback.getContext("2d"),
     ratio       = null,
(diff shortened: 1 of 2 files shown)


BoatsAreRockable:
In terms of functionality this works fantastically! I'll take a proper look at the code later today.

BoatsAreRockable:
Actually more like tomorrow.

app/js/main.js
```diff
@@ -172,41 +172,36 @@ function startup() {
...
-            notifySuccess("Camera successfully connected.");
+    preview.addEventListener("canplay", function() {
+      if (!streaming) {
+        height = preview.videoHeight / (preview.videoWidth / width);
+
+        playback.setAttribute("width", preview.videoWidth.toString());
+        playback.setAttribute("height", preview.videoHeight.toString());
+        streaming = true;
+        ratio = width / height;
+        aspectRatio = ratio.toFixed(2);
```
BoatsAreRockable: I'm aware this was present before you created this branch, but this aspect ratio stuff does not work and should be removed.

app/js/main.js
```diff
@@ -306,7 +301,12 @@ function startup() {
     // Listen for leaving frame rate input
     inputChangeFR.addEventListener("blur", function() {
         inputChangeFR.value = frameRate;
-        if (inputChangeFR.value > 60 || inputChangeFR.value < 1 || NaN || inputChangeFR.length > 2) {
```
BoatsAreRockable: :dizzy_face:

app/js/main.js
```diff
@@ -542,18 +534,18 @@ function takePicture() {
...
-        var data = canvas.toDataURL("image/png");
+        var frame = new Image();
+        frame.src = playback.toDataURL("image/png");
 
         // Store the image data and update the current frame
-        capturedFrames.push(data);
+        capturedFrames.push(frame);
         totalFrames++;
-        console.info(`Captured frame: ${data.slice(100, 120)} There are now: ${totalFrames} frames`);
+        console.info(`Captured frame: ${frame.src.slice(100, 120)} There are now: ${totalFrames} frames`);
```
BoatsAreRockable: Also aware this is legacy code, but the slice part of this console log is no longer necessary. It was initially done to prove to myself that the captured frames were in fact different.

BoatsAreRockable:
This is very good, I shall the merge it. The comments I made relate to things beyond the scope this PR and can be addressed separately.

PR #177, opened 2016-06-26
Compress build.js output folders

BoatsAreRockable:
This PR adds [node-archiver](https://github.com/archiverjs/node-archiver) to build.js allowing the Win32 and OSX64 builds to be compressed to zip archives and the Linux builds to tar.gz archives.

Because the version number was bumped to 0.7.1 at the wrong time and the existence of the confusing v0.7.1-alpha1 build I will release v0.7.2 after this is merged.

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,6 +1,6 @@
 /* Build Boats Animator Executables
 
-Usage: node build <PLATFORMS> <PACKAGES>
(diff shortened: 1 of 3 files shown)


le717:
I hope you are enjoying template strings! :D

BoatsAreRockable:
One can't use too many templates strings! ;)

PR #180, opened 2016-07-04
Support multiple notifications

### LINKED ISSUE #51: Support for multiple notifications
If a notification is shown, and another comes up, it overwrites the current notification. There should either be a window for the notification list, or the notifications can stack on top of each other so when one goes away (maybe by clicking on it?), the other is shown.

le717:
Fixes #51. Finally got around to implementing this. Nothing too special about it. Main issue was getting new notifications to stack nicely, but I got it worked out.

Also reindented a few more scripts/areas according to code style guidelines.

### DIFF
--- a/app/js/file.js
+++ b/app/js/file.js
@@ -149,6 +149,7 @@ module.exports = {};
     });
   }
 
+
   // Public exports
   module.exports.copy = copyFile;
   module.exports.read = readFile;
--- a/app/js/newsfeed.js
+++ b/app/js/newsfeed.js
@@ -148,6 +148,7 @@ module.exports = {};
       });
   }
 
+
   // Public exports
   module.exports.load = load;
 }());
--- a/app/js/notification.js
+++ b/app/js/notification.js
@@ -92,6 +92,7 @@ module.exports = {};
    * @param {String|Number} [msg=""] - The message to display.
    */
   function info(msg) {
+    console.info(`Info: ${msg}`);
     _hideDelayed(_show(_make("info", msg)));
   }
 
@@ -101,6 +102,7 @@ module.exports = {};
    * @param {String|Number} [msg=""] - The message to display.
    */
   function error(msg) {
+    console.error(`Error: ${msg}`);
     _hideDelayed(_show(_make("error", msg)));
   }
 
@@ -110,6 +112,7 @@ module.exports = {};
    * @param {String|Number} [msg=""] - The message to display.
    */
   function success(msg) {
+    console.log(`Success: ${msg}`);
     _hideDelayed(_show(_make("success", msg)));
   }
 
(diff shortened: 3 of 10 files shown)


BoatsAreRockable:
Thanks for implementing this! I'm currently away but hopefully I can find some time to review and test this out in the next few days.

le717:
I hope you like it. As I said, it's not too different from the old system, but different enough that I can't simply push it to `master`. :)

app/js/file.js
```diff
@@ -149,6 +149,7 @@ module.exports = {};
     });
   }
 
+
```
BoatsAreRockable: Nit: the extra line is unnecessary here.

app/js/newsfeed.js
```diff
@@ -148,6 +148,7 @@ module.exports = {};
       });
   }
 
+
```
BoatsAreRockable: Extra line also seems unnecessary here (unless 2 lines before the exports was intentional?)

le717: > (unless 2 lines before the exports was intentional?)

Yes.

app/js/notification.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /**
+   * Hide a notification after a set time.
+   * @see _make
+   *
+   * @param {Object} notify - A notification object.
+   * @returns {Object} The notification object given.
+   */
+  function _hideDelayed(notify) {
```
BoatsAreRockable: I'd suggest that this function and `_show` are reordered to make things easier to follow.

le717: I am not following your reasoning here...

(BTW, `_hideDelayed()` is deliberately named that way. There are reasons. :) )

BoatsAreRockable: @le717 `_make` returns an object to `_show`, which returns an object to `_hideDelayed`.

My thinking is that the order that these functions appear in the code should reflect this order.

BoatsAreRockable:
These are looking very good! The only visual improvement I would suggest is adding a fade out animation when the notifications are hidden.

app/js/notification.js
le717: OK, makes sense. Will reorder ASAP.

BoatsAreRockable: @le717 If you still think the current order is better then I'm happy to leave it how it is. I won't lose sleep over this! ;)

BoatsAreRockable: @le717 missed your reply, please ignore my previous comment!

le717:
Method moved per review. I will look into the fade out. The notification suddenly going away is very deliberate in design. (it was complicated enough just to get it to slide in), but I'll see what I can do. You can always go ahead and merge this if you'd like.

PR #182, opened 2016-07-18
Add a loading animation to the newsfeed

### LINKED ISSUE #170: Loading animation for newsfeed
The newsfeed always takes a few seconds to load so it would be good if some sort of loading animation could show while this happens.

BoatsAreRockable:
This fixes #170 using an adapted version of @le717's [animation](http://codepen.io/le717/pen/yYZOZR).

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -18,7 +18,13 @@
                     <h2><i class="fa fa-newspaper-o"></i>Recent News</h2>
                 </div>
                 <ul>
-                    <li><div class="container-news"></div></li>
+                  <li>
+                    <div class="container-news">
+                      <div class="loading-dots">
+                        <h1 class="dot one">.</h1><h1 class="dot two">.</h1><h1 class="dot three">.</h1>
+                      </div>
+                    </div>
+                  </li>
                 </ul>
             </div>
         </div>
--- a/app/index.html
+++ b/app/index.html
@@ -21,7 +21,7 @@
                   <li>
                     <div class="container-news">
                       <div class="loading-dots">
-                        <h1 class="dot one">.</h1><h1 class="dot two">.</h1><h1 class="dot three">.</h1>
+                        <span class="dot one">.</span><span class="dot two">.</span><span class="dot three">.</span>
                       </div>
                     </div>
                   </li>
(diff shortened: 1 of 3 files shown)


le717:
Even though you `display: none`, the animation is still playing. Either add `animation-play-state: paused` to the hidden class or (possibly the better idea), just `.remove()` the animation from the DOM when done.

app/index.html
```diff
@@ -18,7 +18,13 @@
                     <h2><i class="fa fa-newspaper-o"></i>Recent News</h2>
                 </div>
                 <ul>
-                    <li><div class="container-news"></div></li>
+                  <li>
+                    <div class="container-news">
+                      <div class="loading-dots">
+                        <h1 class="dot one">.</h1><h1 class="dot two">.</h1><h1 class="dot three">.</h1>
```
le717: I don't remember having these as `h1` elements, but I guess I did? `span`s would be more fitting.

BoatsAreRockable: @le717 yeah they were `h1`s. My guess is that you might of done it so you didn't have to change the font size in CSS. Will change to `span`s though.

BoatsAreRockable:
@le717 I've just pushed some changes to this.

app/js/newsfeed.js
```diff
@@ -119,7 +119,7 @@ module.exports = {};
    * @returns {Boolean} Always returns true.
    */
   function displayNews(posts) {
-    loadingDots.classList.add("hidden");
+    loadingDots.parentElement.removeChild(loadingDots);
```
le717: Did you check if `loadingDots.remove()` works

BoatsAreRockable: @le717 sorry I hadn't realised `.remove()` was an actual method! Just committed again using `remove()` which does indeed work. I'd suggest squash merging this particular branch when the time comes.

app/js/newsfeed.js
```diff
@@ -119,7 +119,7 @@ module.exports = {};
    * @returns {Boolean} Always returns true.
    */
   function displayNews(posts) {
-    loadingDots.parentElement.removeChild(loadingDots);
+    loadingDots.remove();
```
le717: `.remove()` is a long-needed shortcut to what you had. See also: https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove

PR #188, opened 2016-10-29
Ensure the save directory exists before using it

### LINKED ISSUE #186: First frame not stored if directory has to be created
If the frame export location is set to a directory that no longer exists and a frame is captured, Boats Animator proceeds to recreate the directory and save captured frames to there. However when this occurs the first frame captured is not saved to this directory.

le717:
Fixes #186. Also corrects the indentation of a few small places.

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -829,42 +829,42 @@ function decodeBase64Image(dataString) {
 /**
  * Save the captured frame to the hard drive.
  *
- * @param {Number} id The frame ID to save.
-*/
+ * @param {Number} - id The frame ID to save.
+ */
 function saveFrame(id) {
-    "use strict";
-    var fileName = "";
+  "use strict";
+  var fileName = "";
 
-    // 1K+ frames have been captured
-    if (id >= 1000) {
-      fileName = `frame_${id}`;
-    }
+  // 1K+ frames have been captured
+  if (id >= 1000) {
+    fileName = `frame_${id}`;
+  }
 
-    // 100 frames have been captured
-    else if (id >= 100) {
-      fileName = `frame_0${id}`;
-    }
+  // 100 frames have been captured
+  else if (id >= 100) {
+    fileName = `frame_0${id}`;
+  }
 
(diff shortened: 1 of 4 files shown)


app/js/main.js
```diff
@@ -139,14 +140,17 @@ function closeAnimator() {
 
 function startup() {
   "use strict";
-  // There is no set save directory, set one
-  if (!_getSaveDirectory()) {
+
+  let saveDirectory = _getSaveDirectory();
+
+  // There is no set save directory or the directory does not exist
+  if (!_checkSaveDirectory(saveDirectory)) {
```
BoatsAreRockable: If this check fails some method of blocking the capturing of frames needs to occur

app/js/main.js
```diff
@@ -754,27 +762,27 @@ function _displaySaveDirectory(dir) {
...
-  return localStorage.getItem("default_directory");
+  return localStorage.getItem("save-directory");
 }
 
 /**
- * Create the frame save directory.
+ * Create the app save directory.
  *
- * @param {String} - The directory to create.
+ * @param {String} dir - The directory to create.
```
BoatsAreRockable: Good spot!

le717:
I would approach it a different way. If no directory is set, the button should be `disabled=true` (or something to that effect), then when a directory is finally set, the button is reenabled. Also ensure the blocking affects the shortcuts. I'd rather disable functionality than add checks like that in places they don't really fit (like `takePicture()`).

app/js/main.js
le717: Heh, yea, I flubbed this last PR... oops

BoatsAreRockable:
Hi @le717, your approach to fixing this sounds good! 😄 Would you like to submit a fix? Or shall I do so?

le717:
I've hopefully fixed everything. Take another look and let me know. :)

BoatsAreRockable:
This does appear to now be fixed. Many thanks! 😄

PR #191, opened 2016-12-20
Make notifications flexible

le717:
The notifications would break if the message was too long. This makes them flexible so the entire message is contained.


BoatsAreRockable:
Seems to do as intended!

PR #192, opened 2016-12-20
Update Font Awesome to 4.7.0

le717:
Just as the title says.


le717:
Self-merging, as this is just a routine library update.

PR #195, opened 2016-12-22
Update to nw.js 0.19.2

le717:
I've been playing with this. Unlike updating to 0.13, this was a easy upgrade, only needing to fix our WebRTC usage. All seems to be stable and working, but I'll let you play with it and merge it when you are happy.


BoatsAreRockable:
I can't see any issues with this 👍

Issue #189, opened 2016-12-19
Refactor html and css


charlielee:
I'd like to refactor the HTML and CSS used in Boats Animator a bit:
* There are currently `div` elements used as `#main-event`, `#sidebar`, and `#statusBar` - these should be moved to their HTML5 semantic equivalents  (`main`, `aside` and `footer`).
* Generally div usage should be reduced, for instance each button in the playback controls is currently a `div` - these should be moved to buttons with `-webkit-appearance: none`.
* Simplify the HTML by potentially moving control panels in `animator.html` from tables use Flexbox.
* Is `#sidebar-content` needed as well as `#sidebar`? 
* There are a number of common elements to the index, about and animator pages. Styles such as sidebar and footer should be moved from being duplicated across 3 different style-sheets to a general `common.css` file.
* When these changes are implemented a significant amount of progress on the switch to 2 spaced indentation (#140) should be made.

BoatsAreRockable:
Update:
* Most of the major semantic element moving and `common.css` work was done in #198 
* Currently I am working on moving a number of different parts of the interface over to flexboxes in the `flexbox` branch.
* I am also further tidying up the html and css.

BoatsAreRockable:
I am shortly due to merge the `flexbox` branch as #208. This will just leave the following point in the OP:
> Generally div usage should be reduced, for instance each button in the playback controls is currently a div - these should be moved to buttons with -webkit-appearance: none.

charlielee:
I think this is low priority and hence can be closed. Reducing div usage is no bad thing but can be done as parts of the program get updates.

PR #198, opened 2017-04-07
Refactor html css #189

BoatsAreRockable:
This brings major improvements to the organisation of Boats Animator's HTML and CSS. From a front-end perspective there should be no visible changes.

### Parts of #189 that have been implemented:
* `#main-event`, `#sidebar`, and `#statusBar` -have been moved to their HTML5 semantic equivalents  (`main`, `aside` and `footer`).
* Removed `#sidebar-content`
* Common elements on the index, about and animator page moved to a general `common.css` file.
* Html files fully 2 spaced now

### Still to do
* Simplify the HTML by potentially moving control panels in `animator.html` from tables to use Flexbox.
* Generally div usage should be reduced further, for instance each button in the playback controls is currently a `div` - these should be moved to buttons with `-webkit-appearance: none`.

**While there's still more refactoring work to do I'd like to merge what I've done so far**


le717:
Just double checking: you're putting this up for review/merge, correct?

BoatsAreRockable:
Yes please!

BoatsAreRockable:
I will be looking to merge this shortly unless there are any major reservations against doing so.

app/js/main.js
```diff
@@ -204,15 +204,6 @@ function startup() {
...
   /* ======= Listeners ======= */
-  // Change resolution
-  document.querySelector("#capture-options li:first-child").addEventListener("click", function() {
-    notification.info("This feature is not yet available!");
-  });
-
-  // Change camera
-  document.querySelector("#capture-options li:last-child").addEventListener("click", function() {
-    notification.info("This feature is not yet available!");
-  });
```
le717: Why did these event listeners get dropped?

BoatsAreRockable: These were removed because the `Change resolution` and `Change camera` options in the sidebar that these listeners refer to were also removed. These in turn were removed because firstly they are just placeholders and secondly because the partially complete WebRTC work for these options will use elements in the UI located elsewhere.

le717:
I can't pull this to look at it until after finals, but if you think it's good, go ahead and merge. My question was answered.

Issue #199, opened 2017-04-07
Developer menu to toggle live reloading


BoatsAreRockable:
It would be helpful if Boats Animator's top menu had a "developer" sub-menu that only displayed when an SDK build of NW.js was run. The main option this would have would be a toggle to allow live reload of the application on changes. This would be very helpful and it would certainly help me speed up development.

For the reloading I would most likely use the snippet on the page below for [chokidar](https://github.com/paulmillr/chokidar):
https://github.com/nwjs/nw.js/wiki/Livereload-nw.js-on-changes
**Edit: turns out someone made a nice package for this. Will probably implement it.**
https://github.com/1j01/nw-dev

The easiest way I can think of to see if an SDK build is being run is to check if the `win.showDevTools()` command exists.

http://stackoverflow.com/questions/40453913/nwjs-is-there-way-to-check-if-useing-the-sdk-version
https://github.com/nwjs/nw.js/issues/5062

**Edit: the checking for win.showDevTools() method does not seem to work unfortunately**

BoatsAreRockable:
Okay so I've decided I'm going to use nw-dev and add it to `package.json` as a `devDependency`.

I can't find a satisfactory way of implementing a check for an SDK build so as a quick fix I'm going to add `toggleLiveReload()` as a function of `utils.js` which you can simply run from the console. This function will add or remove `src="node_modules/nw-dev/lib/dev.js"` to a blank `script` tag with the id `#dev-reload-script` on each of the html pages.

le717:
Implemented in #200.

PR #200, opened 2017-04-07
Add live reloading function to utils.js #199

BoatsAreRockable:
This add a live reloading function (issue #199) to utils.js for development purposes. To activate live reloading `utils.setLiveReload(true)` should be typed into the devtools console. This choice is then remembered for the remainder of the current session.

I initially wanted to do perform a check for a SDK build of NW.js and have a toggle for this feature in the GUI under a Developer menu, however I was unable to find an easy way of doing so. I still think this feature is very useful in console only form though and I'd like to see this implemented asap.

### DIFF
--- a/app/about.html
+++ b/app/about.html
@@ -27,6 +27,7 @@
         </ul>
     </div>
 
+    <script id="dev-reload-script"></script>
     <script src="js/utils.js"></script>
     <script src="js/about.js" async></script>
 </body>
--- a/app/animator.html
+++ b/app/animator.html
@@ -152,6 +152,7 @@
         </ul>
     </div>
 
+  <script id="dev-reload-script"></script>
   <script src="js/utils.js"></script>
   <script src="js/main.js"></script>
 </body>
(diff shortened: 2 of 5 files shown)


BoatsAreRockable:
Hey @le717, it's been a while for both of us!

1. Yes `sessionStorage` is required because when a change is detected the page is basically refreshed causing the state of JavaScript variables to be reset. 

2. This would work in a non-SDK build, although since live reloading is currently triggered manually in the console you'd have to add `utils.setLiveReload(true)` to somewhere in the code.

3. You are correct node_modules is not packaged, however I'm not sure why you'd need live reloading in a packaged build. This works in both SDK/non-SDK builds run without building nw.js (ie by typing `nwb nwbuild -v 0.19.3 -r .` or `nwb nwbuild -v 0.19.3-sdk -r .` in the console).

le717:
That answer works for me. I don't have BA on my computer right now or time to test it, so if it works for you, go ahead and merge. 👍

PR #208, opened 2017-05-19
Remove unnecessary CSS and convert elements to use Flexbox

BoatsAreRockable:
In this PR:
* Tables used for design purposes have been replaced by Flexboxes where possible.
* Unnecessary CSS has been removed
* Move more duplicated CSS to `common.css`
* Increased usage of HTML5 Semantic elements
* `animator.html` is now fully indented with 2 spaces.

### DIFF
--- a/app/about.html
+++ b/app/about.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html>
+
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=2">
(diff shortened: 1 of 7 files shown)


BoatsAreRockable:
@le717 sorry I didn't mean to request a review from you for this! I don't think this particularly needs reviewing. Perhaps a bit of testing, although that is on hold until after my exams finish...

BoatsAreRockable:
Will most likely be shipping this later today. It should also be mentioned that be moving the control panels to use flexbox, the UI scales much better when the window is resized to smaller sizes.

Issue #204, opened 2017-04-19
Handled cameras being plugged in and removed


charlielee:
The program should be able to handle cameras being added or removed. Chromium 57 adds the `ondevicechange` event listener (we are currently on Chromium 55):

https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/ondevicechange

le717:
I really need time to work on the camera code. I could integrate this very easily once the mass of camera code is in place.

BoatsAreRockable:
@le717 I actually started working on the camera code again yesterday and today (see the `webrtc` branch). Currently I've got source changing implemented.

charlielee:
Implemented in #211!

PR #211, opened 2017-12-19
Camera source switching and resolution selection

charlielee:
This feature has been in the works for a quite while now, but it is finally in a state where I am happy to merge it.

Key features:
* User can switch between different camera sources
* A button to "refresh" the source list for sources connected after startup or to "fix" the current source being removed.
* Supported images resolutions are detected and stored for a given camera
* User can switch between the different supported resolutions

@le717 I know it has been a rather long time, but I would greatly appreciate it if you could have a brief look at this, if not then no problem though! 😄

I'd like to merge this asap so I can move ahead with making some progress on Boats Animator over the Christmas holiday period.


le717:
I totally forgot about this. The last semester swept me up. As you can see by my activity, I've only really been able to work in short sprints for months, nothing supportive of work like this. I'll start by congratulating you on implementing this! It's probably the most demanding part of the software so to do it all by yourself is quite a feat and show of dedication!

I'm currently pulling down the code and getting the environment set back up, so I'll take a look at this shortly :)

le717:
Some things:
* Did you research #204? A refresh button is fine but I didn't know if that would do the work for us.
* Removing a camera and refreshing does not remove the camera from the dropdown.
* Changing the resolution always brings up a notification saying the camera has been connected. I am well aware that _technically_ this is how it works but I am curious if there's an easy way to display a "Resolution successfully changed" notification instead.
* Per suggestion of rioforce, the current resolution should be also be displayed in the status bar.
* If I have the camera select or resolution select downdown open and press `<Enter>`, it takes a picture vs. selecting the option. Minor nit but an accessibility issue (one that effected me quickly because I use my keyboard a bit for navigation).
* I've pushed a few minor commits modifying some minor things. I've not really looked carefully at the code itself.

charlielee:
Thanks for having a look so promptly :)

> Did you research #204? A refresh button is fine but I didn't know if that would do the work for us.
I didn't really try this but it should be a couple of lines to trigger the `Camera.enumerateDevices()` method with the `ondevicechange` event so I'll add it. I'll see whether the refresh button should stay as well.

> Removing a camera and refreshing does not remove the camera from the dropdown.
This is actually intentional. It seemed the best way to handle devices being removed temporarily, like when a camera is being moved or gets knocked somehow.

> Changing the resolution always brings up a notification saying the camera has been connected. I am well aware that _technically_ this is how it works but I am curious if there's an easy way to display a "Resolution successfully changed" notification instead.
I'll see how straightforward this is :)

> Per suggestion of rioforce, the current resolution should be also be displayed in the status bar.
Will add!

> If I have the camera select or resolution select downdown open and press `<Enter>`, it takes a picture vs. selecting the option. Minor nit but an accessibility issue (one that effected me quickly because I use my keyboard a bit for navigation).
Good point, I was looking to refactor the keyboard shortcuts after this and move to using 'Mousetrap`. I think I'll save this work for then.

charlielee:
@le717 I've pushed the changes you suggested, except for the issue with `<Enter>` which I'll sort when I redo the shortcuts.

> Did you research #204? A refresh button is fine but I didn't know if that would do the work for us.

It seems the reason I didn't do this before was because we were on NW.js v0.19.3 (Chromium 55) and the `ondevicechange` event was added with Chromium 57. I've bumped us up to NW.js v0.26.6 (Chromium 62) now and implemented this. :)

le717:
I've pushed a few more tweak commits. I've yet to really review the code because I want to work out the usability first.

All looks good _except_ not removing a disconnected camera from the dropdown. This is causing a couple of bugs, best explained in a video: https://youtu.be/40kBExTw37M

Basically, if I unplug the currently active camera (when the input goes black), select another camera, then select the now-removed camera, it'll load the last successful camera but the UI will show all elements of the removed camera. I showed this to rioforce and even he thought the unplugged camera was active. The threat of a webcam becoming unplugged, is his evaluation, is minor and not significant enough to justify this behavior. I still recommend removing the camera from the list. It looks like you'll want the [`ended`](https://developer.mozilla.org/en-US/docs/Web/Events/ended) event to detect this. When removed, an info notification saying "Camera disconnected" should be shown.

charlielee:
Obviously the behaviour in the video is a bug, my intention was for a black screen to be displayed when a removed camera is selected. But I can see how not removing disconnected devices is confusing, so I'll change it so that disconnected cameras are removed from the dropdown. :)

Although, I must say it will be difficult for me to test this because I accidentally left my external webcam at University. I had to use my microphone to test `ondevicechange`! 😆

> Also, is there any particular reason you didn't update nw.js to 0.27.2 but stopped at 0.26.6?

This was an oversight on my part, however I have just tried updating to NW.js 0.27.2 and the main "animator" window does not load, with just a white screen appearing instead?! 😟

le717:
I'll take a look at it. I have two USB webcams plus the one on my laptop at my disposal xD

le717:
I've been working on this all afternoon (I'm off my coding game...). It's been surprisingly tricky because it's creating unintended side effects :\ Hopefully I can get this wrapped up quickly

le717:
I've pushed what changes I have + a few cleanup commits (I've still not really looked over the code fully).

If you're able to test this code (maybe you can borrow a webcam?), you'll find the following behavior.

The problem is the `devicechange` event is shared for device adding/removing, which makes sense (`ended` only fires when the app is closing, I checked) but complicates matters. We can update the camera list all we want but the resolutions selection never changes. Further, the only remaining camera (for me) gets selected by the browser but the `changed` event never fires and since it's already "selected", I can't click it. I have to click the refresh button to get it to switch to the camera. If I just click the refresh button in code (or extract the event for the refresh button and reuse it for `devicechange`) then we get `change` on the camera select firing twice if unplugging or the current camera getting the resolution twice if plugging in. There's also an occasional "Cannot find a camera to use" error before actually finding the camera (so both notifications are visible at once, sometimes. Sometimes I can trigger it repeatedly by selecting a resolution before it kicks). Triggering the `change` event in the code produces yet again different behavior. It's very weird and I've not been be able to find a side-effect free method all afternoon/tonight. Reusing the code for the refresh button gets us the closest to where it should be (it even falls back to a working camera) but the side-effects are yucky. Note all this only happens if the current camera goes away. Removing a source with a different camera selected is good. No side-effects there.

In short: 😐

charlielee:
I should be able to get hold of a webcam from my Dad's work but that won't be until the new year. It's frustrating that fixing this behaviour has been so troublesome, we're so close to finishing this as well! 😢😅

le717:
I know right?? It's really sad. :(

BTW, once this gets merged, the next release is 0.8.0. :P

le717:
Just came across this:
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/tWzutytXsqc/lGaWCFdHAgAJ

I checked and we are affected by the change. It won't affect us directly right now but it'd be better to go ahead and make this change.

Also, any update on getting a webcam so we have another pair of eyes seeing the weird behavior?

charlielee:
Thanks for alerting me to that change! It might as well as be part of the PR.

I'm reunited with my webcam again, however I have exams for the next 2 weeks. I'll try and get round to investigating this for myself as soon as I can :)

charlielee:
@le717 I know it has been a very long time, but I'd like to dissect the issue you had. I don't really know how to fix the problem you've described and what we should do.

Maybe this sounds bad, but how likely are users to unplug their camera? Or would it help if camera source changing was not done "live"? ie done in a seperate settings window

charlielee:
@rioforce sorry for the delay in replying. I can't believe this PR has been open for nearly a year! This project is always on the back of mind, and something I really want to get going again (maybe this Christmas will finally be the time for a big push...).

In relation to this PR, I can confirm resolution and source changing are implemented, and work reliably. 😄 The issue is in how cameras being unplugged is handled.

I think the best solution is to make source and resolution occur in a separate window rather than in the sidebar. No guarantees, but I have just broken up from uni for the holidays so I'll see if I can implement that in the next few days. Would be happy to merge this at that point.

---

On a side note, we really need to fix how existing frames in a directory are overwritten if the user doesn't specifically choose a different directory on startup. That bugs me a lot! I know implementing a full projects feature sorts this, but in the short term something is needed.

Finally, I'd really like to refactor a lot of the code in Boats Animator - I've learnt a lot in the last 2 years! I think the very intiutive UI and potential power of Boats Animator (for free!) makes it still very worthwhile pursuing development, however long it takes me!

charlielee:
Update: so I have decided to keep source/resolution changing in the sidebar. I have improved the handling of devices been disconnected so that "no device selected" shows if the currently selected device is the one disconnected. The refresh button has also been removed.

I'm happy that

le717:
> I have improved the handling of devices been disconnected so that "no device selected" shows if the currently selected device is the one disconnected. The refresh button has also been removed.

This was literally the one thing that prevented my OK for merge. I'm glad to hear it's fixed! :D

TBH, because this is such a large PR that's long overdue (and admittedly, huge), once it's rebased with master, I'm trusting your judgement for merging and we can deal with any refactoring/cleanup after the fact. Besides, we probably need to update nw.js and I think the webrtc api has been updated since this was started...

le717:
Now just be sure to tell BiM about the update 😉

PR #216, opened 2017-12-26
Switch to live view after playback stops

### LINKED ISSUE #214: Playback should end with live view rather than the last frame captured
This is the behaviour implemented in the majority of other stop motion animation programs I have used.

le717:
Fixes #214.

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -232,7 +232,6 @@ function startup() {
 
   // Play/pause the preview
   btnPlayPause.addEventListener("click", function() {
-    // Make sure we have frames to play back
     if (totalFrames > 0) {
       (isPlaying ? videoPause : previewCapturedFrames)();
     }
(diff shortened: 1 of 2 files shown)


le717:
It's not as much as the last frame not being played as much as it being played first. I counted to 5 on my fingers and slowed the playback to 2 or 3 FPS. The playback order was 5, 1, 2, 3, 4. This also repos on `master` but slightly differently: 5, 1, 2, 3, 4, 5. I'm looking at it now but it's surprisingly a difficult little bug to track doen

le717:
@charlielee OK, this should be fixed. The last frame should now be displayed. I did not fix the last frame displaying first because that's a different bug and it'll be better to see if it still occurs in #211 and fix it there.

charlielee:
Thanks for fixing this 😃 I wonder how long the "5, 1, 2, 3, 4, 5" bug has been present... 😕

Issue #82, opened 2015-11-15
Revise confirm dialog behavior


le717:
Spun off https://github.com/BoatsAreRockable/animator/pull/70#issuecomment-152787716.

> The `confirm()` API was intentionally written to return true or false depending on the button clicked because it is generic and supports many uses cases. While currently our use case is a simple "if yes, run function else do nothing", thinking long term this may not be the case and this code will have to be refactored to support it. I am thinking it would be better to refactor this to copy the generic behavior of `confirm()` to support future use cases.

BoatsAreRockable:
Adding to this, the "OK" button in confirm dialogs should be focused by default - allowing the user to press enter to select it.

BoatsAreRockable:
Another improvement that should be made to confirm dialogues is that the HTML tabindex method be used to switch between the OK and cancel buttons:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

charlielee:
This is linked to the introduction of a new pop window class (issue #213)

charlielee:
Implemented in #229

PR #229, opened 2018-12-26
Implement revised confirm dialogues #82

charlielee:
This PR implements revised confirm dialogues using the SweetAlert library rather than the previous custom solution. While the old dialogues matched the general theme of BA better, I have chosen to leave the new ones as is, because they provide a greater contrast with the background.

The new dialogues return a promise and are configurable with all of the library's features, so are much more customisable than before.

## Before:

screenshot 100

## After:

screenshot 101

### DIFF
--- a/app/js/main.js
+++ b/app/js/main.js
@@ -82,6 +82,7 @@ var streaming = false,
   notification = require("./js/notification"),
   saveDirectory = require("./js/savedirectory"),
   menubar = require("./js/menubar"),
+  swal = require("./lib/sweetalert"),
 
   // Sidebar
   dirChooseDialog = document.querySelector("#chooseDirectory"),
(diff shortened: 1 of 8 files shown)


app/js/main.js
```diff
@@ -800,59 +806,35 @@ function saveFrame(id) {
...
+ *                    Resolves true if confirm was selected and null if the alert was dismissed.
  */
-function confirmSet(callback, args, msg) {
+function confirmSet(swalArgs) {
   "use strict";
-  confirmText.innerHTML = msg;
-  confirmContainer.classList.remove("hidden");
-  btnConfirmOK.focus();
+  // Set default SweetAlert argument values
+  swalArgs.title = ("title" in swalArgs) ? swalArgs.title : "Confirm";
```
le717: `("title" in swawlArgs)` is more of a Python convention. Since you're only accessing a top-level prop, change it and the other lines to this format:
```suggestion
  swalArgs.title = swalArgs.title ? swalArgs.title : "Confirm";
```

charlielee:
@le717 thank you for reviewing this and your continued support of this project! 😄

Issue #227, opened 2018-12-23
Store Camera.list in local storage


charlielee:
I think we should store the supported resolutions of detected cameras between sessions, as the device media ID associated with a camera seem to stay the same between sessions. 

This would mean that the resolution selection loading screen would only need to happen the first time each camera is detected, rather the the first time every session. The last resolution used could also be stored.

Cameras with no resolutions detected (due to an error or the user never selecting them) would not be stored between sessions.

The eventual preferences screen could have #101 could have a "Reset Device List" option.

charlielee:
I'm currently working on this. Once this and #224 are implemented, I'll release v0.8.1.

le717:
I'm not actually sure if localSession is persisted across app launches. I know I store a cached version of the newsfeed in it but I don't actually know if it is persistent. Do a quick check to confirm it is persistent before going further

charlielee:
Can confirm that `localStorage` is persistent, while `sessionStorage` isn't :smile:

le717:
Completed by #230.

PR #230, opened 2019-01-05
Store Camera.list in local storage #227

charlielee:
This stores found cameras and resolutions to localStorage. This means that the program doesn't need to calculate which resolutions a camera supports on every single launch.

### DIFF
--- a/app/js/camera.js
+++ b/app/js/camera.js
@@ -36,7 +36,6 @@ module.exports = {};
     this.name = name;
     this.curResolution = null;
     this.resolutions = resolutions;
-    this.responsive = null;
   }
 
   /** Instance methods */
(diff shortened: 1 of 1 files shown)


charlielee:
> Q: why was this removed?

The commit should really read remove cam "detected" notification, not connected. A notification still displays on successfully connecting to a camera.

I removed the notification because it only shows the when a camera is detected that has never been seen before (ie isn't in localStorage). Even prior to this PR, the notification only showed the first time a camera was connected, and not a second time if you unplugged and replugged it in.

I could amend the behaviour, to notify every time a camera is detected, but I don't think it's that necessary and would require a bunch of checks on every `ondevicechange` event.

Issue #235, opened 2019-03-25
Refactor code to be modular


charlielee:
I have started majorly refactoring the code into modules in the `modules` branch. The code base of this program has got to a point where it is highly difficult to add new features because of it's state.

charlielee:
This may change, but I'm hoping to end up with a structure similar to the following:
```
/src (renamed app directory)
  - /app.js
  - /ui
    - /component1
      - /component1.css
      - /component1.js
    - /component2
      - /component2.css
      - /component2.js
  - /main
    - /component1
      - /component1.css
      - /component1.js
    - /component2
      - /component2.css
      - /component2.js
- /common
  - /common.css
  - /common.js
    
```

* `/ui` will contain components to do with visual elements of the program (eg frame reel, status bar, onion skinning window, capture window, playback window)
* `/main` will contain components to do with the logic of the program eg the state of the animation (frames captured/current frame, webrtc, frame file creation etc, save directory)

charlielee:
My plan is to do this in at least 2 stages.

* Phase 1: modularise ui related js and css
* Phase 2: modularise core logic (ie capture and playback)

Issue #86, opened 2015-11-16
Project system


le717:
Another brain dump incoming. As always, anything is up for discussion.
- [JSON-based](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON)
- Ends in Boats-unique extension for (at least on Windows) shell registration (click the file and Boats opens it)
- Contains following details
  - Project name
  - Description
  - Author name
  - Creation date
  - Frame save path (default: `/frames`)
  - FPS
  - Number of captured frames
- Absolute paths are not stored to enable portability (just copy the project files and use it anywhere)
- Able to frames outside project folder (state such is not recommended to user)
- If required to create external path, info notification saying so
- Anything else I'm missing (I'll edit this issue as needed)

rioforce:
Solid idea, but it is missing takes. Often, someone will want to use the same project and capture more than one take, shot, or even scene without making a new project. Thus, the main project should be a folder, inside should be a folder for each take (which would include the frames, and if a specific json was needed for each take, the frames would be in a subfolder).

BoatsAreRockable:
https://github.com/nwjs/nw.js/wiki/Handling-files-and-arguments

BoatsAreRockable:
https://support.microsoft.com/en-us/kb/184082

https://msdn.microsoft.com/en-us/library/cc144158%28VS.85%29.aspx?f=255&MSPPError=-2147217396

le717:
Dude, I've created file associations with Inno Setup before, [it is really easy](https://github.com/le717/PatchIt/blob/rewrite/Tools/Windows/PiInstaller.iss#L120-L125). :)

BoatsAreRockable:
Oh sweet :) what about Mac/Linux though? Is this something we could only do once we have a dmg/deb/rpm setup for them?

le717:
There's only been a Mac in my house for a few months because dad bought one (and I've never used it), and Linux has only been used to recover from computer disasters. I'm sure it is possible, but I haven't the slightest clue. :P

charlielee:
Closing this issue as I have recently began splitting it out into several others

Issue #238, opened 2019-04-10
Improve playback performance


charlielee:
* Load blank `div` onto framereel while the thumbnail blob is still being rendered to make things seem more instant
*  An option to playback in a lower resolution than captured

charlielee:
Canvas optimizations: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas

charlielee:
An interesting api i found: https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

charlielee:
In my testing, capture performance is now mostly satisfactory. Playback performance needs work though. In particular the first run through of playback is slow

PR #239, opened 2019-04-13
Modularisation part 2

charlielee:
This provides a major step towards making BA's core code more modular #235. While multiple smaller PRs would be preferable to this humongous one, it was simply easier to do all of this in one go.

`main.js` has reduced from 733 to 431 lines and been split into:

* AudioManager
* ConfirmDialog
* Project
* Take
* FrameRate

The concept of "Projects" and "Takes" now exists, however no additional functionality has been added yet (for now internally a Project with one Take is created on start up). The classes have just been refactored in such a way to make it much more straightforward to add a full Project system #86 later on.

This leaves the following parts of `main.js` to refactor later:

* New window creation
* Playback
* WebRTC
* Takeframe / undo frame event listeners

I am very excited for what this refactor has done to the codebase. Things are so much better to work with now, and I anticipate adding new features will be soo much easier!


le717:
Unfortunately, I don't have the time to go through everything and give a proper review (my GitHub activity right now is partially hobby and partially school!) because of college and other things, but I have been trying to keep an eye on the commits as you have pushed them. The only thing I'd suggest, and this is so you can continue to improve your JS skills, is to learn about [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) and [`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const). They are nice and can help make coding easier. I personally don't use `var` anymore unless I absolutely have to.

I'm really happy to see these changes! It definitely will help with future improvements. But there's a question: does all this help with the speed issues reported?

charlielee:
> The only thing I'd suggest, and this is so you can continue to improve your JS skills, is to learn about [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) and [`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).

You raise a very good point, I really should start doing that! :) I won't switch to `let`/`const` in this PR put I'll do a big move away from `var` soon.

> But there's a question: does all this help with the speed issues reported?

In general in this refactor I have focused more on reorganisation than optimisation (for now) so generally there hasn't been much speed improvement.

However in #236 I switched from frame capture from `toDataUrl` to `toBlob` which caused a notable capture performance improvement. :) I have started #238 to note any other optimisations I can think of.

Issue #221, opened 2018-12-21
Icon not been added to executable when building Windows binaries


charlielee:

charlielee:
Moved to v0.9.x as I'll relook at build methods then

charlielee:
Fixed in #249

PR #249, opened 2019-08-02
Change build to use nwjs-packager

charlielee:
This PR changes the build process to use the https://github.com/charlielee/nwjs-packager module I have developed. `nwjs-packager` builds the app with `nw-builder` and then outputs packages of the user's choice (currently .zip, .tar.gz and Inno Setup .exe).

This brings a bunch of benefits:
* `build.js` doesn't need to contain custom code to make archives and packages
* Adding .deb/.rpm/.dmg support can be done independently of boats animator
* Moves us off the defunct nwjs-builder
* Fixes windows icon issue #221

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ node_modules/
 jsconfig.json
 *.~is
 npm-debug.log
+cache/
+build/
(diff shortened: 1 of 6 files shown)


le717:
I opened my email to 12 emails from here. I'm good with that though, because progress is good 👍

Issue #215, opened 2017-12-26
Short play option


charlielee:
There should be the ability to playback just the last few frames captured.  How many this is should be set in the eventual Application Preferences System #101

charlielee:
Implemented in #257

PR #257, opened 2020-05-05
Implemented Short Play 

AndreasDemenagas:
Implementation of issue #215 -- Having the ability to short play (preview) the last couple of frames of the animation. 

Currently set to preview the last 5 frames.


charlielee:
Thanks for your contribution @AndreasDemenagas! :) I will review this in the next few days

app/ui/PlaybackControls/PlaybackControls.js
```diff
@@ -18,13 +19,22 @@
         global.projectInst.playback.toggleVideoLoop();
       });
 
+      // Short Play 
+      // Currently set to preview last 5 frames of the animation
+      btnShortPlay.addEventListener("click", function() {
+        if (takeInst.getTotalFrames() > 0) {
```
charlielee, line 25: Add a check for !global.projectInst.playback.isPlaying

app/animator.html
```diff
@@ -77,6 +77,9 @@ <h2>Select a camera source to begin!</h2>
           <div id="btn-frame-last" title="Last frame" class="button">
             <i class="fa fa-fast-forward"></i>
           </div>
+          <div id="btn-short-play" title="Short Play" class="button">
+            <i class="fa fa-sort"></i>
```
charlielee, line 81: Use the `fa-play-circle` icon (unless you have a better suggestion)

app/main/Playback/Playback.js
```diff
@@ -92,12 +92,15 @@
...
+     * @param {Integer} number - Frame number that previewing begins
      */
-    previewCapturedFrames() {
+    previewCapturedFrames(number) {
+      console.log("Previewing from frame " + number);  
       // Display playback window
       if (global.projectInst.getCurrentMode() === "capture") {
         global.projectInst.setCurrentMode("playback");
-        this.curPlayFrame = 0;
+        this.curPlayFrame = number;
```
charlielee, line 103: Set the curPlayFrame in both capture and playback modes to fix the playback mode issue mentioned in the longer comment

charlielee: Although i'm not sure if this will break other app functionality. So will need to be tested

AndreasDemenagas: Testing the above, I saw that setting curPlayFrame in both capture and playback mode it disables the ability to play from the selected frame when pressing the normal play button. If for example I have frame 13 selected and I click the play button, the entire animation is played backed and not the animation starting in frame 13 as it would have.

charlielee: I think a possible solution is to call the `_displayFrame(id)` function before calling `previewCapturedFrames()` when short play is activated. `_displayFrame` is the function called when a user selects frame from the framereel.

This should mean you won't need to add an id argument to `previewCapturedFrames()`. 

I must admit the playback code for Boats Animator is a bit strange!

charlielee: Also would you be able to make the shortPlay code a method in Playback.js? Then it can be reused when a keyboard shortcut or menu bar item is added

AndreasDemenagas: > Also would you be able to make the shortPlay code a method in Playback.js? Then it can be reused when a keyboard shortcut or menu bar item is added

You mean write a different method instead of using `previewCapturedFrames()`?  (Since previewCapturedFrames is already defined in Playback.js)

AndreasDemenagas:
After some thought I decided to move the short play button to the left container, next to the undo frame button. This eliminates the issue you described above, as this container is automatically hidden in playback mode. 

I have also defaulted the `previewCapturedFrames()` method's `number` param to 0, so to not cause confusion when called from the normal play button.  

> **Playback mode:** in playback mode pressing short play causes all frames from the currently displayed frame to be played.

app/main/Playback/Playback.js
charlielee: I mean change:

```javascript
btnShortPlay.addEventListener("click", function() {
  if (takeInst.getTotalFrames() > 0) {
          if (!global.projectInst.playback.isPlaying) {
            const number = takeInst.getTotalFrames() - 5; 
            global.projectInst.playback.previewCapturedFrames(number);
          } else {
            console.log("already playing...");
          }
        }
});
```

To:

```javascript
btnShortPlay.addEventListener("click", function() {
  global.projectInst.playback.shortPlay();
});
```

and then in Playback.js add:

```javascript
shortPlay() {
  if (takeInst.getTotalFrames() > 0) {
          if (!global.projectInst.playback.isPlaying) {
            const number = takeInst.getTotalFrames() - 5; 
            global.projectInst.playback.previewCapturedFrames(number);
          } else {
            console.log("already playing...");
          }
        }
}
```

(some variable names will probably need changing here)

charlielee:
This looks ready to merge now, thanks for your help! 😄

Issue #253, opened 2020-01-18
Camera feed not loading on macOS Catalina


charlielee:
The camera feed does not seem to be working on macOS 10.14 or 10.15. A bit of a major issue!

I think the issue is that the camera permission request dialogue is not triggered. See:
https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos

Possibly some solutions listed here:
https://stackoverflow.com/questions/56084303/opencv-command-line-app-cant-access-camera-under-macos-mojave

Probably need to mess about with the generated plist file. Probably something to handle within https://github.com/charlielee/nwjs-packager

charlielee:
Electron handles this nicely it seems:
https://electronjs.org/docs/api/system-preferences#systempreferencesaskformediaaccessmediatype-macos

A related issue:
https://github.com/electron/electron/issues/15985

I'm struggling to find anything similar for nodejs

charlielee:
Maybe related to https://github.com/nwjs/nw.js/issues/7229

charlielee:
Maybe this'll help https://github.com/gch1p/mojave-permissions

arudnev:
I just stumbled upon this ticket when I realized that we had similar issue for a while.
Not sure how it worked for us on macOS Mojave, but after we added notarization support for app to work on macOS Catalina we noticed that camera and microphone permissions are not requested.
Adding missing entitlements as part of codesign process seems to resolve the issue.

https://github.com/nwjs/nw.js/issues/7330

So, something like that as entitlements file should fix it

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>com.apple.security.device.camera</key>
    <true/>
    <key>com.apple.security.device.audio-input</key>
    <true/>
    <key>com.apple.security.device.microphone</key>
    <true/>
    <key>com.apple.security.automation.apple-events</key>
    <true/>
    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
    <true/>
    <key>com.apple.security.cs.allow-jit</key>
    <true/>
    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
    <true/>
    <key>com.apple.security.cs.disable-executable-page-protection</key>
    <true/>
[...]

Issue #203, opened 2017-04-19
Export to video


charlielee:
Eventually the ability to export a project's captured frames to video should exist. At the moment when playback occurs a canvas animation occurs and one way of turning this into a video would be using https://github.com/antimatter15/whammy, although this would only allow WebM videos to be exported.

A more flexible way I can see us eventually considering would be using a combination of two different Web APIs that both seem to be pretty experimental:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/captureStream
https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/MediaRecorder

BoatsAreRockable:
If we need used Whammy, Handbrake-js could then be used to encode the `WebM` video to other video formats. https://github.com/75lb/handbrake-js

Btw I should empathise that this feature is a very low priority right now. Individual frames output is perfectly adequate since most editing programs support image sequence importing.

charlielee:
Recently I came across this JavaScript port of FFmpeg which could used for the video export function:
https://github.com/Kagami/ffmpeg.js

Whammy seems to not be actively maintained and FFmpeg.js is far more versatile and even supports things such as Web Workers. Here's an example of FFmpeg converting an image sequence to video:
http://hamelot.io/visualization/using-ffmpeg-to-convert-a-set-of-images-into-a-video/

DerJacques:
Great project! 
Did you by any chance stumble upon [Videoshow](https://www.npmjs.com/package/videoshow) during your research?  
I think it would be perfect for boats-animator. The only thing I'm not certain about is how easy it is for you to include ffmpeg in your application. That being said, you would probably experience the same issue using ffmpeg.js.

charlielee:
> Great project!

@DerJacques thank you very much, although there is still much to do!

> Did you by any chance stumble upon Videoshow during your research?
I think it would be perfect for boats-animator.

No I haven't seen Videoshow before, thanks for pointing it out! 😄 I have just been experimenting with it and it definitely seems very promising. It certainly removes a lot of the complexity of using FFmpeg. So far, my only concern with it is that each image of the video has a duration set in seconds rather than frames. Since each frame in stop motion is very short I am concerned that this could lead to dropped frames in the output video. Also it could lead to longer videos being slightly longer or shorter than expected

> The only thing I'm not certain about is how easy it is for you to include ffmpeg in your application. That being said, you would probably experience the same issue using ffmpeg.js.

[...]

PR #258, opened 2020-05-05
Export video

charlielee:
This PR isn't ready for release yet. But I figured I might as well open it to give visibility to what is currently being worked on.

The biggest hold up has been having to write an entire nwjs packaging library due to nw-builder's shortcomings, and no currently available alternative.

New features:

- [x] Implement video export via ffmpeg #203
- [x] Upgrade to [nwjs-packager](https://github.com/charlielee/nwjs-packager) 1.x to support running `npm install --production` and fix numerous macOS issues #253
- [x] Test export feature on macOS
- [x] Fix export dialog being closeable during the export
- [x] Allow users to specify custom ffmpeg arguments 
- [x] Check if output file already exists to stop overwrite
- [x] Text export feature on Windows
- [x] Test export feature on Linux


charlielee:
This is so close now! Just need to add a check that the video file isn't being overwritten

charlielee:
I've noticed that videos exported with the default options seem to have difficulty opening in some media players (other than VLC, but what do you expect! 😂). I will need to poke around with the ffmpeg options

AndreasDemenagas:
> I've noticed that videos exported with the default options seem to have difficulty opening in some media players (other than VLC, but what do you expect! 😂). I will need to poke around with the ffmpeg options

I have been able to reproduce this issue on macOS Catalina (10.15.6) using QuickTime Player.

charlielee:
The QuickTime issue was due to the `crf` parameter being set to 0. I don't fully understand it but the docs describe "a subjective sane range is 17-28", so set to 17 I have!
https://trac.ffmpeg.org/wiki/Encode/H.264

If there are no more bugs, then this branch is now ready!

Issue #219, opened 2018-01-14
Remember the size and position of windows and restore them in that place after app relaunch


charlielee:

charlielee:
Should be able to achieve this with the id option of window.open https://docs.nwjs.io/en/latest/References/Window/#windowopenurl-options-callback

AndreasDemenagas:
I have a solution almost ready. Should I move forward with a PR?

charlielee:
Yes please :)

PR #263, opened 2020-05-10
Remember size and position of window 

AndreasDemenagas:
This PR implements issue #219 

(Please ignore the terrible brach name. I was supposed to write issue-219-rem.size 😂) 

Using a similar approach, the index.html page (welcome window) could be added an id property to remember its size as well. (Although I think we should leave it like that, since we wouldn't want it to occupy the entire screen after reopening - even if the user scales it to do so). Hopefully that makes sense.

### DIFF
--- a/app/ui/WindowManager/WindowManager.js
+++ b/app/ui/WindowManager/WindowManager.js
@@ -7,11 +7,7 @@
 
   class WindowManager {
     static setListeners() {
-      // Maximise window
-
-      // This call seems to bypass the id property in the window.open() options 
-      //win.maximize();
-
+      
       /**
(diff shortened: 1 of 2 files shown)


app/ui/WindowManager/WindowManager.js
```diff
@@ -8,7 +8,9 @@
   class WindowManager {
     static setListeners() {
       // Maximise window
-      win.maximize();
+
+      // This call seems to bypass the id property in the window.open() options 
+      //win.maximize();
```
charlielee, line 13: Yeah this can go, please delete these commented out lines :)

AndreasDemenagas:
Just deleted them : )

Issue #272, opened 2020-10-10
Bug when exporting video after deleting frames


charlielee:

charlielee:
Have documented some manual tests. Everything seems to be working when running the app from source. Will need to build an installer and try again

### Test 1

1. Capture 6 frames
2. Press "undo last frame" once
3. Video export OK

### Test 2

1. Capture 6 frames
2. Delete frame 4
3. Video export OK

### Test 3

1. Capture 6 frames
2. Delete frame 4
3. Capture 2 more frames
4. Video export OK

### Test 4

1. Capture 8 frames
2. Delete frame 4
3. Delete the new frame 4
4. Capture 2 more frames
5. Video export OK

charlielee:
I've found the bug, it's kind of weird to reproduce 😎 
Unlikely to affect most users because it's when you try to capture frames in a directory that already has frames.

The current way that frames could be accidentally overwritten is terrible tbh. Until we have "takes" implemented i'm going to write a method to use the current unix time on start up in the file names of all frames.

PR #273, opened 2020-10-21
Add unique id to each take

charlielee:
This adds a unique id to each take generated from the current Linux epoch time. I also made frame ids 5 digits long rather than 4

This will fix issue #272 and one of the most unintuitive "features" of the app where frames will be overwritten if you don't choose a new directory each time you start the app

It's true that technically if a user were to create 2 projects in the same second, or change their PCs clock back frames could still be overwritten with this change. However this is highly unlikely and this will do until a project system is fully implemented.

### DIFF
--- a/app/main/ExportVideo/ExportVideo.js
+++ b/app/main/ExportVideo/ExportVideo.js
@@ -46,7 +46,7 @@
         <br>
         <select id="presetSelect">
           <option value="veryslow">Very slow</option>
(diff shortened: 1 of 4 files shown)


charlielee:
Tested on Windows ✔️

charlielee:
Tested on macOS ✔️

charlielee:
Tested on Ubuntu ✔️

Issue #274, opened 2020-10-24
Move to Electron


charlielee:
This project should be moved to Electron. Some of the main reasons I wish to move are:

* NW.js is missing a solid build/packaging library
* NW.js doesn't properly implement camera permissions, making macOS support feel slightly hacky (there is not an equivalent to Electron's [systemPreferences](https://www.electronjs.org/docs/api/system-preferences#systempreferencesgetmediaaccessstatusmediatype-windows-macos) api)
* Electron seems better supported and to have a much larger developer eco-system

Over the years of this project's development Electron has become far more popular and feature filled than NW.js. It has a large the eco system of packages and other tools around it that we could take advantage of. Whereas NW.js doesn't even have a reliable build library. I have developed the [nwjs-packager](https://github.com/charlielee/nwjs-packager) library but it requires a lot more work to be even close to something like https://github.com/electron-userland/electron-builder.

I would rather spend my free time developing this app, then working on build tools. This will be a big job but I think it will be worth it.

I will start experimenting with how this will work soon. Other features such as projects can be developed normally still and we will have to work out how to port them to the Electron branch

AndreasDemenagas:
I completely agree with moving the project to electron. It is true that it is way more developed than nwjs and I believe it can really help the project. 

Also, the refactoring of the code into modules that you have done should really help make it easier to make an electron build.

charlielee:
NW.js specific features we currently use:

#### nwdirectory

Adding the [`nwdirectory`](https://nwjs.readthedocs.io/en/latest/References/File%20Dialogs/#featured-extended-by-nwjs) attribute to the "choose directory" option allows the app to get the full path of a chosen directory

#### Menu

The `Menu`/`MenuItem`/`nw.Window.get().menu` options are used to generate the top menu bar shown.

https://nwjs.readthedocs.io/en/latest/References/Menu/

#### Shell

`nw.Shell.openExternal(url)` is used to open links from the news section, about window and help menu.

https://nwjs.readthedocs.io/en/latest/References/Shell/

#### Window

The window

charlielee:
Currently working on this with a very basic electron and electron-builder setup. Lots of boilerplates/tutorials suggest adding things like webpacker, react or testing suites. These are interesting things to consider for the future but outside of this issue's scope

PR #282, opened 2020-12-27
Move to Electron

charlielee:
This reworks the app to use Electron rather than NW.js #274. It will be a major PR once complete.


le717:
How much more still needs to be done before this is complete? I don't think I have my USB webcam but I can test things out using my built-in laptop webcam.

charlielee:
> How much more still needs to be done before this is complete?

@le717 I'd say it's about 90% complete now. The main things left are:

* Reimplement dialog box for selecting export video directory (selecting "Export video" will error at the moment)
* Some CSS fixes related to the general window layout (probably flexbox related)
* Restore keyboard shortcuts showing in the menu bar
* Test/configure electron-builder correctly
* Fix window dimensions not being stored when you choose the "Main Menu" menu bar option

I am hoping to finish very soon as I'm not working next week.

AndreasDemenagas:
One thing I noticed while testing in macOS Big Sur is that the Menu Bar skips the first item "File" and puts in straight into the Menu Item of the application name. (See attached picture). 

This is probably because in macOS the first menu item is always reserved by the system and holds the name of the application. 

Checking if the platform is Darwin and if so skipping the first array item should fix this. 

Screenshot 2020-12-28 at 2 35 03 PM

le717:
I'm going to recommend that once this is merged, tag it as 0.11.0, release it as a beta (also check the pre-release box in Releases), and ask people to download and use it. Don't add too many other changes. It's going to be easier having many others find issues related just to this and fix them for a next release than trying to find everything now.

charlielee:
> One thing I noticed while testing in macOS Big Sur is that the Menu Bar skips the first item "File" and puts all the rows that should be in "File" straight into the Menu Item of the application name. (See attached picture).
> 
> This is probably because in macOS the first menu item is always reserved by the system and holds the name of the application.

@AndreasDemenagas Thanks for spotting this. I haven't tested on macOS/Linux yet so let me know if see anything else.

> I'm going to recommend that once this is merged, tag it as 0.11.0, release it as a beta (also check the pre-release box in Releases)

@le717  Yep, will bump to 0.11.0 after this is merged.

Also export video *should* work now. Note file paths with a space will fail to export as #278 is not merged into this branch yet.

charlielee:
- [ ] Check possible bug with confirm take
- [ ] Finalise and test packages on Linux
- [ ] Finalise and test packages on macOS
- [ ] Finalise and test packages on Windows

charlielee:
This is finally ready to go! 🚀 Hopefully I can actually work on some features again now.

PR #562, opened 2025-09-25
Folder schema loading

ikyman:
Clicking on the "Open Project" opens up the folder dialogue. This does not yet open at the project directory, instead opening at the . From That, you can select which project to load. Absent a selection screen to choose which take to be loaded, take 1 is automatically loaded. 

ProjectFilesContextProvider is now always available, even from the startup screen; The capture screen required both the Project and Take to be loaded in react redux, and ProjectFilesContextProvider is the module used for loading the project and take from memory. Therefore, it had to be available from the startup screen.

I marked the "useProjectDirectory" hook as depricated. Waiting for Dexie to trigger is just not fast enough to reliably have the right projectDirectory at the right time. This is now tracked manually, through a projectDirectory useState.

Takes store their name, so they can refer to their respective folders.


charlielee:
Sorry for the delay in reviewing this. I will try and look this weekend 😊

